PR #225, opened 2021-02-03
Avoid temp seed DB persistence - Use recoil for temporary state and flush it after backup

eddycpc:
### Problem

There's a scenario where the user might close the app just before completing the backup.
At that stage, the wallet is still persisted in plain text because wallet storage encryption only happens after the user has backup. 
In that scenario, the wallet can still be found in plain text in the indexedDB files

### Old Behaviour 

During wallet creation user flow, it used to happen in 2 steps :

Step 1: Persist on-disk(DB) temporary wallet data, here they were not encrypted yet since it still had to be used later for users to  backup

Step 2: When the backup is completed, the previously persisted wallet data was finally encrypted ( Its seed phrase )

### New Behaviour & Solution

Here there's no more temporary DB or disk persistence. It's now being kept in `recoil/memory` temporary storage and flushed afterward.

So now during creation UI flow, wallet creation only happens in 1 step compared to the previous 2 steps creation.

### DIFF
--- a/src/models/Wallet.ts
+++ b/src/models/Wallet.ts
@@ -52,6 +52,11 @@ export interface CustomConfigFormValue {
   indexingUrl: string;
 }
 
+export function setPhrase(wallet: Wallet, phrase: string): Wallet {
+  wallet.encryptedPhrase = phrase;
+  return wallet;
+}
+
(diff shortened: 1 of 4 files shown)


crypto-matto (approved):
Works fine 👍🏻

leejw51crypto:
rebasing now

PR #233, opened 2021-02-06
Handle max transaction amount (sub fee value) - Validate against minimum value - Show fee in preview

eddycpc:
Screen Shot 2021-02-05 at 11 30 34 AM
Screen Shot 2021-02-05 at 7 10 29 PM


leejw51crypto (approved):
lgtm

leejw51crypto:
how about limiting fee?
if it's too high, ask twice?

eddycpc:
> how about limiting fee?
> if it's too high, ask twice?

We are currently using a fixed fee for now

PR #276, opened 2021-02-25
Fix regression on background job - Update price API BaseURL

eddycpc:
Fixes https://github.com/crypto-com/chain-desktop-wallet/issues/275

### DIFF
--- a/src/service/WalletService.ts
+++ b/src/service/WalletService.ts
@@ -200,7 +200,8 @@ class WalletService {
   public async syncAll(currentSession: Session | null = null) {
     const session =
       currentSession == null ? await this.storageService.retrieveCurrentSession() : currentSession;
-    if (!currentSession) {
+
+    if (!session) {
       return;
     }
     // Stop background fetch tasks if the wallet configuration network is not live yet
--- a/src/service/WalletService.ts
+++ b/src/service/WalletService.ts
@@ -197,19 +197,22 @@ class WalletService {
     return broadCastResult;
   }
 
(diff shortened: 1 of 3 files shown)


leejw51crypto (approved):
lgtm

src/service/WalletService.ts
```diff
@@ -200,7 +200,8 @@ class WalletService {
   public async syncAll(currentSession: Session | null = null) {
     const session =
       currentSession == null ? await this.storageService.retrieveCurrentSession() : currentSession;
-    if (!currentSession) {
+
```
leejw51crypto, line 203: === null ?

leejw51crypto: for readibilty
(currentSession === null) ?  await... : currentSession?

src/service/rpc/MarketApi.ts
```diff
@@ -11,7 +11,7 @@ export class CroMarketApi implements IMarketApi {
 
   constructor() {
     this.axiosClient = axios.create({
-      baseURL: 'https://chain.crypto.com/api',
+      baseURL: 'https://crypto.org/api',
```
leejw51crypto, line 14: this is ok,
how about using one file containing all const, and definitions ?
static config?

src/service/WalletService.ts
eddycpc: > === null ?

I use == because it checks for both null and undefined which I'm more interested in

https://stackoverflow.com/a/20688019

calvinaco (approved):

PR #283, opened 2021-03-02
Problem: ledger not working after unplug & plug (fix #282)

### LINKED ISSUE #282: Problem: ledger not working after unplug & plug
further tx signing is not working after unplug & plug

leejw51crypto:
Solution: connect, disconnect for each operation


eddycpc (approved):
LGTM

PR #294, opened 2021-03-10
Add backend logic support for wallet deletion

eddycpc:


src/storage/StorageService.ts
```diff
@@ -30,6 +30,10 @@ export class StorageService {
     );
   }
 
+  public async deleteWallet(walletID: string) {
+    return this.db.walletStore.remove({ identifier: walletID }, { multi: true });
```
leejw51crypto, line 34: await? like deleteWallet above

leejw51crypto (approved):
lgtm

crypto-matto (approved):

src/storage/StorageService.ts
eddycpc: Await is already implicit in that scenario.

https://eslint.org/docs/rules/no-return-await

PR #296, opened 2021-03-12
Fix vulnerability warning - react-dev-utils

eddycpc:


crypto-matto (approved):
Shall we stick with 11.0.4?

eddycpc:
> Shall we stick with 11.0.4?

The lib is safe from 11.0.4 upward

Issue #302, opened 2021-03-22
Problem : Support re-delegation UI implementation


eddycpc:

Nawaz-Malik143:
Keep it up

eddycpc:
Been handled

PR #304, opened 2021-03-22
Add support for Redelegation : backend & UI/UX

eddycpc:
Screen Shot 2021-03-22 at 6 28 59 PM
Screen Shot 2021-03-22 at 6 29 11 PM

Fixes  https://github.com/crypto-com/chain-desktop-wallet/issues/302

Fixes https://github.com/crypto-com/chain-desktop-wallet/issues/301

### DIFF
--- a/src/config/StaticConfig.ts
+++ b/src/config/StaticConfig.ts
@@ -25,7 +25,7 @@ const TestNetConfig: WalletConfig = {
   derivationPath: "m/44'/1'/0'/0/0",
   explorerUrl: 'https://crypto.org/explorer',
   indexingUrl: 'https://crypto.org/explorer/api/v1/',
-  nodeUrl: 'https://testnet-croeseid.crypto.org',
+  nodeUrl: CroNetwork.Testnet.defaultNodeUrl,
   network: CroNetwork.Testnet,
 };
 
@@ -37,7 +37,7 @@ const MainNetConfig: WalletConfig = {
   enabled: true,
   name: 'MAINNET',
   derivationPath: "m/44'/394'/0'/0/0",
-  nodeUrl: NOT_KNOWN_YET_VALUE,
+  nodeUrl: CroNetwork.Mainnet.defaultNodeUrl,
   explorerUrl: NOT_KNOWN_YET_VALUE,
   indexingUrl: NOT_KNOWN_YET_VALUE,
   network: CroNetwork.Mainnet,
--- a/src/config/StaticConfig.ts
+++ b/src/config/StaticConfig.ts
@@ -24,7 +24,7 @@ const TestNetConfig: WalletConfig = {
   name: 'TESTNET',
   derivationPath: "m/44'/1'/0'/0/0",
   explorerUrl: 'https://crypto.org/explorer',
-  indexingUrl: 'https://crypto.org/explorer/api/v1/',
+  indexingUrl: 'https://crypto.org/explorer/croeseid/api/v1/',
   nodeUrl: CroNetwork.Testnet.defaultNodeUrl,
   network: CroNetwork.Testnet,
 };
@@ -101,3 +101,4 @@ export type Network = {
 };
 
 export const FIXED_DEFAULT_FEE = 5_000;
+export const FIXED_DEFAULT_GAS_LIMIT = 300_000;
--- a/src/pages/home/components/UndelegateFormComponent.tsx
+++ b/src/pages/home/components/UndelegateFormComponent.tsx
@@ -2,59 +2,63 @@ import { FormInstance } from 'antd/lib/form';
 import { Alert, Form, InputNumber } from 'antd';
 import React from 'react';
 import { Session } from '../../../models/Session';
+import { TransactionUtils } from '../../../utils/TransactionUtils';
 
 export const UndelegateFormComponent = (props: {
   currentSession: Session;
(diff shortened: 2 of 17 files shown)


crypto-matto:
I also added Top Validators auto complete

crypto-matto (approved):

src/service/WalletService.ts
```diff
@@ -197,6 +179,50 @@ class WalletService {
     return broadCastResult;
   }
 
+  public async sendReDelegateTransaction(
```
leejw51crypto, line 182: how about
sendRedelegateTransaction ?

src/pages/home/components/UndelegateFormComponent.tsx
```diff
@@ -0,0 +1,60 @@
...
+          label="Undelegate Amount"
+          validateFirst
+          rules={[
+            { required: true, message: 'Undelegate amount is required' },
+            {
+              pattern: /[^0]+/,
+              message: 'Undelegate amount cannot be 0',
+            },
+            {
+              max: Number(props.undelegateFormValues.undelegateAmount),
```
calvinaco, line 42: I wonder if this will have some precision lose if the amount is too big or have floating points?

calvinaco (approved):
LGTM, I left a small comment.

src/pages/home/components/UndelegateFormComponent.tsx
eddycpc: This being used only for form validator which expects only number|Number types for max

Also this won't interfere on the precision because it's only for validating maximum form input value, it's not being used to modify underlying input numbers that we care about.

calvinaco: Yes, but I mean it may still be affected by the precision? Take an example of string `'1000000000.12345678` or `111111111.12345671`, if it passed to Number it will be wrong.

So this will affect the max number a user can accept. In fact the `<InputNumber />` will also treat number as number in JavaScript.

There is a `stringMode` in `antd` which is designed for money
https://ant.design/components/input-number/#components-input-number-demo-digit

calvinaco: Thanks to @crypto-matto who discovered this when we are working on another project. I think we should consider using it for desktop wallet as well.

crypto-matto: @calvinaco Fixed the issue by referencing 
https://github.com/crypto-com/token-migration-webtool/pull/104

src/service/WalletService.ts
eddycpc: Not sure what you mean here btw.

crypto-matto:
Fixed #308 
Fixed precision issues on max & min validation on all transaction amount input fields:
`Send`
`Delegate`
`Un-Delegate`
`Re-Delegate`

Nawaz-Malik143 (commented):
Kya khe rhey hain sab

src/pages/send/send.tsx
```diff
@@ -118,16 +118,27 @@ const FormSend = () => {
...
     walletAsset,
     AddressType.USER,
   );
   const customAmountValidator = TransactionUtils.validTransactionAmountValidator();
+  const customMaxValidator = TransactionUtils.maxValidator(
+    maximumSendAmount,
+    'Sending amount exceeds your available wallet balance',
+  );
+  const customMinValidator = TransactionUtils.minValidator(
+    fromScientificNotation(currentMinAssetAmount),
```
Nawaz-Malik143, line 136: What are you saying

src/pages/send/send.tsx
```diff
@@ -118,16 +118,27 @@ const FormSend = () => {
...
+  const scaleUpBalance = scaledBalance(walletAsset); // From BaseXYZ balance to XYZ balance
+  const currentMinAssetAmount = getCurrentMinAssetAmount(walletAsset);
+  const maximumSendAmount = scaleUpBalance;
+
   const customAddressValidator = TransactionUtils.addressValidator(
     currentSession,
     walletAsset,
     AddressType.USER,
   );
   const customAmountValidator = TransactionUtils.validTransactionAmountValidator();
```
Nawaz-Malik143, line 126: $10000

Nawaz-Malik143 (commented):
Complete

src/utils/TransactionUtils.ts
```diff
@@ -45,6 +46,30 @@ export class TransactionUtils {
...
+        if (!Big(value).gt(max)) {
+          return Promise.resolve();
+        }
+        // eslint-disable-next-line prefer-promise-reject-errors
+        return Promise.reject(reason);
+      },
+    });
+  }
+
+  public static minValidator(min: string, reason: string) {
```
Nawaz-Malik143, line 61: - [ ]

src/config/StaticConfig.ts
```diff
@@ -24,7 +24,7 @@ const TestNetConfig: WalletConfig = {
   name: 'TESTNET',
   derivationPath: "m/44'/1'/0'/0/0",
   explorerUrl: 'https://crypto.org/explorer',
-  indexingUrl: 'https://crypto.org/explorer/api/v1/',
```
Nawaz-Malik143, line 27: - [ ]

Nawaz-Malik143 (commented):
> > - [ ] #- [x] [**```~~~~

```**]()

PR #310, opened 2021-03-24
Problem: Ledger does not accept custom derivation path (fix #309)

### LINKED ISSUE #309: Problem: Ledger does not accept custom derivation path
If one use the Devnet with custom derivation path, the address derived is still the default derivation path.

leejw51crypto:
Add address index field

add address index

remove alert

remove console.log


eddycpc:
Any screenshots of the UI changes ?

eddycpc:
Is the idea of having the custom index supposed to work on ledger only or do we plan to have it working on both ledger and the normal wallets types ?

Nawaz-Malik143 (dismissed):

src/pages/create/create.tsx
```diff
@@ -432,6 +433,24 @@ const FormCreate: React.FC<FormCreateProps> = props => {
...
+          },
+          {
+            pattern: /^0$|^[1-9][0-9]?$|^100$/,
+            message: 'Index between 0 - 100',
+          },
+        ]}
+        hidden={props.isWalletSelectFieldDisable}
+      >
+        <Input placeholder="0" defaultValue="0" />
+      </Form.Item>
```
eddycpc, line 453: I think this should generally be optional and it should only show if the user is willing to provide a custom derivation index.

cc @crypto-matto @leejw51crypto

eddycpc (changes requested):

leejw51crypto:
image

src/pages/create/create.tsx
crypto-matto: If so then remove
`initialValue="0"` & `defaultValue="0"`
```
{
            required: true,
            message: 'Please input your address index',
},
```
@leejw51crypto can it auto fill in 0 index at backend when `addressIndex` is empty?

leejw51crypto: let's make it visible only for ledger

eddycpc: Another question I asked at the top, do we plan to have this working only for Ledger type of wallets?

Do we have the specs requirements to support this feature on normal wallets ?

leejw51crypto: only for ledger

leejw51crypto: for normal wallets, it can be useful, but can be done as another issue.

eddycpc: That's good @leejw51crypto let's have it first working for Ledger as you mentioned. Cool cool!

linfeng-crypto (approved):

linfeng-crypto:
ZEMU test need fix too

leejw51crypto:
ok, fixing now

leejw51crypto:
ready to merge

eddycpc (approved):

eddycpc (approved):

calvinaco (approved):

PR #315, opened 2021-03-24
Hide memo for the moment

crypto-matto:


leejw51crypto (approved):
ok

src/pages/send/send.tsx
```diff
@@ -189,10 +189,9 @@ const FormSend = () => {
           </div>
         </div>
       </div>
-      <Form.Item name="memo" label="Memo (Optional)">
+      <Form.Item name="memo" label="Memo (Optional)" hidden>
```
eddycpc, line 192: This would prevent users to deposit to exchanges ...

eddycpc (changes requested):

src/pages/send/send.tsx
leejw51crypto: how about specifying fee explicitly?

Issue #322, opened 2021-03-25
Problem: missing field of networkFee in old database


leejw51crypto:
add legacy support of networkFee

eddycpc:
Was fixed on 0.1.1 release

PR #323, opened 2021-03-25
Problem: missing field of networkFee in old database #322

leejw51crypto:
Solution: add legacy support

### DIFF
--- a/src/pages/settings/settings.tsx
+++ b/src/pages/settings/settings.tsx
@@ -11,6 +11,8 @@ import { SettingsDataUpdate } from '../../models/Wallet';
 import { Session } from '../../models/Session';
 import ModalPopup from '../../components/ModalPopup/ModalPopup';
 
+import { FIXED_DEFAULT_FEE, FIXED_DEFAULT_GAS_LIMIT } from '../../config/StaticConfig';
+
 const { Header, Content, Footer } = Layout;
 const { TabPane } = Tabs;
 const layout = {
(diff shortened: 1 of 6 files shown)


crypto-matto (commented):
螢幕截圖 2021-03-25 上午10 25 46

We have to change here as well. `settings.tsx`
```
const onFinish = async values => {
    if (
      defaultSettings.nodeUrl === values.nodeUrl &&
      defaultSettings.indexingUrl === values.indexingUrl &&
      defaultSettings.network.chainId === values.chainId &&
      defaultSettings.fee.gasLimit === values.gasLimit &&
      defaultSettings.fee.networkFee === values.networkFee
    ) {
      // No value was updated, we stop here
      return;
    }
```

src/pages/settings/settings.tsx
```diff
@@ -107,27 +109,42 @@ const FormSettings = () => {
...
-        gasLimit: defaultSettings.fee.gasLimit,
+        networkFee,
+        gasLimit,
       });
     }
   }, [form, defaultSettings]);
 
   const onFinish = async values => {
+    const defaultGasLimit =
+      defaultSettings.fee !== undefined ? defaultSettings.fee.gasLimit : FIXED_DEFAULT_GAS_LIMIT;
```
linfeng-crypto, line 138: may be we can use a function getFee and getGasLimit, and not put it into defaultSettings

leejw51crypto: ok

leejw51crypto: https://github.com/crypto-com/chain-desktop-wallet/issues/324

i made another issue

src/pages/settings/settings.tsx
```diff
@@ -107,27 +109,42 @@ const FormSettings = () => {
...
   const history = useHistory();
+  let networkFee = parseInt(FIXED_DEFAULT_FEE, 10);
+  let gasLimit = parseInt(FIXED_DEFAULT_GAS_LIMIT, 10);
 
   useEffect(() => {
     if (!didMountRef.current) {
       didMountRef.current = true;
+
+      if (defaultSettings.fee !== undefined) {
+        networkFee = parseInt(defaultSettings.fee.networkFee, 10);
```
eddycpc, line 120: This value might be bigger than int, I don't think the parseInt is necessary

eddycpc (changes requested):

src/pages/settings/settings.tsx
leejw51crypto: ok, i'll change it

eddycpc (approved):
Looks great!

For people with old DBs, please confirm to us that this has now been resolved 🙇

crypto-matto:
> Looks great!
> 
> For people with old DBs, please confirm to us that this has now been resolved 🙇

I tested with old DB. Working good now.

Issue #325, opened 2021-03-25
Problem: Suggested top validator list not updated with correct network when switching between mainnet & testnet wallet


crypto-matto:

eddycpc:
Should be working on this

PR #327, opened 2021-03-25
Fetch a new validators set when switching wallet

eddycpc:
Fixes https://github.com/crypto-com/chain-desktop-wallet/issues/325

Screen Shot 2021-03-25 at 8 44 29 AM
Screen Shot 2021-03-25 at 8 45 09 AM


crypto-matto (approved):
Problem solved 👍🏻🙌🏻

PR #346, opened 2021-03-27
Problem: adding console.log for tx failure error (fix #345)

### LINKED ISSUE #345: Problem: adding console.log for tx failure error

leejw51crypto:
Solution:add debug log

### DIFF
--- a/src/service/rpc/NodeRpcService.ts
+++ b/src/service/rpc/NodeRpcService.ts
@@ -62,6 +62,7 @@ export class NodeRpcService implements INodeRpcService {
     try {
       const signedBytes = Bytes.fromHexString(signedTxHex).toUint8Array();
       const broadcastResponse = await this.client.broadcastTx(signedBytes);
+      // eslint-disable-next-line no-console
       console.log(`tx failure= ${JSON.stringify(broadcastResponse)}`);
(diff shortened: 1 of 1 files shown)


src/service/rpc/NodeRpcService.ts
```diff
@@ -62,6 +62,8 @@ export class NodeRpcService implements INodeRpcService {
     try {
       const signedBytes = Bytes.fromHexString(signedTxHex).toUint8Array();
       const broadcastResponse = await this.client.broadcastTx(signedBytes);
+      // eslint-disable-next-line no-console
+      console.log(`tx failure= ${JSON.stringify(broadcastResponse)}`);
```
eddycpc, line 66: I'm wondering if there's a reason would we want to console log in release builds? 

The console tab is being hidden in release builds, so even if it logs we can't see the logs

eddycpc (changes requested):
Left a comment

src/service/rpc/NodeRpcService.ts
leejw51crypto: ok, i see

leejw51crypto:
replace this issue with this one
https://github.com/crypto-com/chain-desktop-wallet/issues/349

PR #354, opened 2021-03-29
[Problem Fix #341] Make Mainnet as default selected wallet on creation and imports

### LINKED ISSUE #341: [Problem] Make Mainnet as default selected wallet on creation and imports

crypto-matto:


src/pages/create/create.tsx
```diff
@@ -456,6 +456,7 @@ const FormCreate: React.FC<FormCreateProps> = props => {
           placeholder="Select wallet network"
           onChange={onNetworkChange}
           disabled={props.isNetworkSelectFieldDisable}
+          defaultValue="MAINNET"
```
eddycpc, line 459: Doing this way usually creates warnings as ant design request default values to be put in the form `initialValues` prop

eddycpc (changes requested):

eddycpc (approved):

PR #373, opened 2021-03-31
Persist validators set and load more necessary validator property data ( Backend implementation )

eddycpc:
Fixes https://github.com/crypto-com/chain-desktop-wallet/issues/369


chewy917:
0x93a62da5a14c80f265abc077fcee437bla0efde

crypto-matto:
Shall we keep the top validators in autocomplete for the moment?

crypto-matto (approved):

eddycpc:
> Shall we keep the top validators in autocomplete for the moment?

It's still there supported

PR #375, opened 2021-04-01
Problem (Fix #358) : Before relegations show warning info - Explaining to users how re-delegation really works

### LINKED ISSUE #358: Problem : Before relegations show warning info - Explaining to users how re-delegation really works

crypto-matto:
螢幕截圖 2021-04-01 下午5 34 47


eddycpc (approved):

eddycpc:
This should probably be part of the wallet config.network in the future among other static configuration values

PR #376, opened 2021-04-01
Add default desktop wallet client memo value when no memo is set (Staking transactions types)

eddycpc:
Fixes https://github.com/crypto-com/chain-desktop-wallet/issues/372


crypto-matto (dismissed):
LGTM

PR #381, opened 2021-04-01
Dependency vulnerability warning fix - is-svg old version

eddycpc:


calvinaco (approved):
LGTM.

on the other hand, does the upstream dependency we are usingupdated with the non-vulnerable `is-svg` yet? If they have upgraded does it mean we can remove it from the resolutions list?

crypto-matto (approved):

eddycpc:
> LGTM.
> 
> on the other hand, does the upstream dependency we are using updated with the non-vulnerable `is-svg` yet? If they have upgraded does it mean we can remove it from the resolutions list?

Upstream hasn't released yet a version with a fix last time checked

PR #382, opened 2021-04-01
Problem: Validator list has non-bonded validators

cryptodotbzh:
Hi, the function doesn't fetch all the validators, only 100 mixed with Unbonded and Jailed validators.
So the Top Validators List is incorrect. 
My fix is to filter validators by their status = BOND_STATUS_BONDED

I hope I'm doing thing right, I'm a newbie in GitHub :)


calvinaco (approved):
LGTM, thanks for spotting the issue and your contribution.

PR #385, opened 2021-04-02
Problem: Validator list shows inactive validators

### LINKED ISSUE #383: Problem: Validator list shows inactive validators

calvinlauyh:
Solution: (Fix #383) Filter inactive validators away


src/service/rpc/NodeRpcService.ts
```diff
@@ -157,25 +169,70 @@ export class NodeRpcService implements INodeRpcService {
...
+      (pubKeyMap, validator) => {
+        pubKeyMap[validator.value] = true;
+        return pubKeyMap;
+      },
+      {},
     );
 
-    return response.data.validators
+    return validators
+      .filter(v => v.status === 'BOND_STATUS_BONDED')
```
eddycpc, line 195: We could have probably kept this filter on the top-level URL so that we don't load a large list of JSON objects that is mainly filled with UNBONDED validators.

This way we apply the filter before even loading and everything we load is mostly clean

Issue #393, opened 2021-04-02
Problem: Error [big.js] Invalid number when loading the home page


calvinaco:
## How to reproduce?

- Clear the wallet storage
- Initialize the wallet with password and create a wallet
- Errors displayed when router to the home page

![image](https://user-images.githubusercontent.com/38898718/113430217-5a463100-940c-11eb-9bc6-c8d8bc1591da.png)

eddycpc:
Working on this

Issue #389, opened 2021-04-02
Delegate funds page - top 20 validators list isn't the top 20


chamsters:
The list shown of top validators isn't the top 20- I see numerous in top 20 missing and some in 30s+ showing up. What's the query here?

Issue #384, opened 2021-04-02
Problem: Only top 20 validators are listed


Fafnabch:
Only a selection of up to 20 best validators for selection
There is a maximum of 100 POOLs. It would be Fair Play to see them all.
This way, the Rich will be richer, and those who seek decentralization must fight for survival.

Fafnabch:
And I'll add when I turn on my wallet, I don't have the current status. I welcome the old values of cro. I have to have two wallets and switching between them will show the correct status.

PR #394, opened 2021-04-02
Load and persist validators on network basis - and lots more improvements

eddycpc:
- Load and persist validators on the  network basis, not wallet
- Remove current limit on 20 validators to max 100 validators
- Show total balances in proper currency view via numeral
- Fixed case where null market asset price would crash the app

Screen Shot 2021-04-02 at 7 26 36 PM
Screen Shot 2021-04-02 at 7 31 13 PM

Fixes https://github.com/crypto-com/chain-desktop-wallet/issues/393

Fixes https://github.com/crypto-com/chain-desktop-wallet/issues/389

Fixes https://github.com/crypto-com/chain-desktop-wallet/issues/384


crypto-matto:
Add ellipsis in validator name & website to avoid variable height with up & down pagination buttons
螢幕截圖 2021-04-03 上午2 01 36

crypto-matto (dismissed):
LGTM

PR #397, opened 2021-04-02
v0.1.4 Release : Majorly improving by using correct filter for validator set

eddycpc:


calvinaco:
@eddycpc maybe I have overlooked something. I saw you removed the `fetchAndSetNewValidators` call but I couldn't see it being called in other places?

calvinaco (approved):

Issue #396, opened 2021-04-02
[Feature request] Staking Rewards screen should show the Reward Amount in fiat


Techboy:
**Feature request**.

It would be useful if the Staking Rewards screen showed the total Reward Amount in fiat.

This could be displayed in a section at the top of the screen, like how Total Balance is listed in the Home screen.

eddycpc:
This is a fair request, @crypto-matto can look into this

crypto-matto:
Feature developed in #401

PR #401, opened 2021-04-03
[Feature request #396] Staking Rewards screen should show the Reward Amount in fiat

crypto-matto:
螢幕截圖 2021-04-03 下午11 05 14
螢幕截圖 2021-04-03 下午11 05 31

Fixes https://github.com/crypto-com/chain-desktop-wallet/issues/396


src/pages/staking/staking.tsx
```diff
@@ -527,20 +529,29 @@ const FormWithdrawStakingReward = () => {
...
+  const convertToTabularData = (
+    allRewards: RewardTransaction[],
+    currentAsset: UserAsset,
+    currentMarketPrice: AssetMarketPrice,
+  ) => {
     return allRewards
       .filter(reward => Big(reward.amount).gte(Big(0)))
       .map(reward => {
         const rewardAmount = getUIDynamicAmount(reward.amount, currentAsset);
+        const marketPrice = new Big(currentMarketPrice.price);
```
eddycpc, line 548: On scenario where `currentMarketPrice.price` is null this would crash this view. We need to handle this case.

`currentMarketPrice` can be null or empty '', this would happen on app storage clearing, before new market data has been load.

Similar to this case I fixed yesteray https://github.com/crypto-com/chain-desktop-wallet/issues/393

eddycpc (changes requested):

eddycpc (approved):

Issue #423, opened 2021-04-09
Problem : Add a settings option for configurations changes to apply on all wallets in the app


eddycpc:
**Is your feature request related to a problem? Please describe.**

Wallet users want to have the possibility to change all wallets config like nodeUrl in a single config change

**Describe the solution you'd like**

We can add an opt-in option ( a check box ) that would allow all configurations to replicate across wallets on the same network.

PR #430, opened 2021-04-13
General config update propagation support added

eddycpc:
Fixes https://github.com/crypto-com/chain-desktop-wallet/issues/423


crypto-matto (approved):
LGTM, easy to use

PR #444, opened 2021-04-15
Problem: no expert mode display is incorrect  (fix #443)

### LINKED ISSUE #443: Problem: no expert mode display is incorrect
only display when expert mode is necessary

leejw51crypto:
Solution: show `no expertmode message` correctly

### DIFF
--- a/src/pages/create/create.tsx
+++ b/src/pages/create/create.tsx
@@ -22,7 +22,7 @@ import SuccessCheckmark from '../../components/SuccessCheckmark/SuccessCheckmark
 import IconLedger from '../../svg/IconLedger';
 import {
   createLedgerDevice,
-  detectLedgerStatus,
+  detectConditionsError,
   LEDGER_WALLET_TYPE,
   NORMAL_WALLET_TYPE,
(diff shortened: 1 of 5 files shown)


eddycpc (approved):

calvinaco:
@leejw51crypto can you describe the problem and how it this PR solves it? So far it seems that it only checks for the error "Condition not met" only.

leejw51crypto:
yes,  previously, 
when tx is rejected,  it also display "no expert mode".
because detectLedgerStatus returns "NORMAL".

in new approach, when tx is rejected, "no expert mode" message is not displayed.
and for real "no expert mode", 
```
catch (e) {
      if (walletType === LEDGER_WALLET_TYPE) {
        setLedgerIsExpertMode(detectConditionsError(e.toString()));
      }
```
error starts with`Conditions not satisfied`, and display `no expert mode` correctly.
also in new approach, after catching first error, no more calls ledger api.

leejw51crypto:
previously)
image

now)
image

calvinaco (approved):

PR #447, opened 2021-04-16
Problem (Fix #446): No default sorting on wallet page

### LINKED ISSUE #446: Problem: No default sorting on wallet page
Every time I switch a wallet, the list goes back to it's original "random" list, so if I want to switch wallets again, I need to sort them again, then look for the wallet I need. Would it be possible to for the list to remain sorted? Or at least use alphabetical sorting on the name as default?

crypto-matto:
Default sort wallet name with ascending order


eddycpc:
I remember there used to be a sorting based on current selected wallet, do we still have the current selected wallet on top ?

crypto-matto:
> I remember there used to be a sorting based on current selected wallet, do we still have the current selected wallet on top ?

Since the sorting is on the frontend side, I think that would require a dedicated fixed area, outside of the table, in order to do so.

crypto-matto:
Fixed row for current selected wallet & extracted from wallet list. 
螢幕截圖 2021-04-19 下午5 52 31

crypto-matto:
Fixed row for current selected wallet & extracted from wallet list.

螢幕截圖 2021-04-19 下午7 19 36

ytbigvinny288 (approved):

ytbigvinny288 (approved):

ytbigvinny288 (approved):

ytbigvinny288 (approved):

eddycpc:
The icon on the Ledger wallet type looks asymmetric comparing to the normal wallet.

- I think if we add icon for one wallet type we should also add the icon to the other type

src/pages/wallet/wallet.tsx
```diff
@@ -70,66 +91,115 @@ function WalletPage() {
...
+            ) : (
+              NORMAL_WALLET_TYPE
+            ),
+          dataIndex: 'walletType',
+          // Same as title above
+          render: walletType =>
+            walletType && walletType.length > 2 ? (
+              <>
+                {walletType.charAt(0).toUpperCase() + walletType.slice(1)}{' '}
+                {walletType === LEDGER_WALLET_TYPE ? <Icon component={IconLedger} /> : ''}
```
eddycpc, line 152: If we add an icon for Ledger let's also add an icon for the normal wallet type, that way we keep things symmetrical

eddycpc (changes requested):

crypto-matto:
螢幕截圖 2021-04-27 下午6 12 48

eddycpc (approved):

PR #454, opened 2021-04-19
(Fix #451) Spin loading in front of the home wallet & optimised DB fetching mechanism

### LINKED ISSUE #451: [UX] When wallet home is loading - Don't hide the all wallet home view - We can have the spin loading in front of the home wallet
We can have the spinner loading while having the wallet home visible in the background, currently, the wallet home just becomes white, which doesn't look pretty good.

Screen Shot 2021-04-17 at 1 33 24 PM

We can have the spin in front of the content, an example on here https://ant.design/components/spin/

crypto-matto:
This PR also optimised DB fetching mechanism on all pages. 
螢幕截圖 2021-04-19 下午4 30 25

### DIFF
--- a/src/layouts/home/home.less
+++ b/src/layouts/home/home.less
@@ -122,8 +122,12 @@
     height: 100vh;
     margin-left: @sider-width;
     > .ant-spin-nested-loading :not(.ant-spin-container) .ant-spin {
-      height: 100vh !important;
       max-height: none !important;
+    }
+    &.loading {
+      .ant-spin {
+        height: 100vh !important;
+      }
     }
   }
 }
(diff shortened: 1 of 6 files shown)


eddycpc:
Great work @crypto-matto 

I see mounting checks being removed in this PR, can you confirm that this doesn't bring new warnings about memory leak issues

crypto-matto:
> Great work @crypto-matto
> 
> I see mounting checks being removed in this PR, can you confirm that this doesn't bring new warnings about memory leak issues

The memory leak issue happens when there are potential infinite rendering on components. 

For this PR update, the fetching which causes component to re-render is limited to only when the General State `fetchingDBState` is updated. You may see inside all `useEffect()`, it tracks `fetchingDB` and will be triggered only if it got updated. So the memory leak warning issues is not happening. 

You may checkout the branch and see if you can find any occasions where the warning would appear, and so far it's fine.

eddycpc (approved):

PR #517, opened 2021-06-09
Transfer NFT UI Flow with optimized layout

crypto-matto:
Flow: 
1. Transfer NFT Form
2. Ask for App Password
3. Confirm Transfer NFT Detail
4. Success / Error Popup
螢幕截圖 2021-06-09 下午12 33 42
螢幕截圖 2021-06-09 下午12 34 03
螢幕截圖 2021-06-09 下午12 34 18
螢幕截圖 2021-06-09 下午12 35 20


crypto-matto:
Added NFT Transaction on Home Page
螢幕截圖 2021-06-10 上午2 36 03

src/pages/home/home.tsx
```diff
@@ -137,6 +153,46 @@ function convertTransfers(
...
 
+function convertNftTransfers(allTransfers: NftAccountTransactionData[]) {
+  function getStatus(transfer: NftAccountTransactionData) {
+    if (transfer.success) {
+      return TransactionStatus.SUCCESS;
+    }
+    return TransactionStatus.FAILED;
+  }
+  function getType(transfer: NftAccountTransactionData) {
+    if (transfer.messageType === 'MsgIssueDenom') {
```
eddycpc, line 164: Should we use an enum or constants for this, instead of checking with in-place string 'MsgIssueDenom'

eddycpc (changes requested):

src/pages/nft/nft.tsx
```diff
@@ -169,6 +111,96 @@ const NftPage = () => {
...
+
+      setIsNftModalVisible(false);
+      setIsNftTransferModalVisible(false);
+      setIsNftTransferConfirmVisible(false);
+      setConfirmLoading(false);
+      setIsSuccessModalVisible(true);
+      setInputPasswordVisible(false);
+      const currentWalletAsset = await walletService.retrieveDefaultWalletAsset(currentSession);
+      setWalletAsset(currentWalletAsset);
+
```
eddycpc, line 186: I think after a successful NFT transfer, we can call `walletService.retrieveNFTs()` to update the account NFT  recoil state.

> This way if an NFT was sent to another address, it won't keep being shown again in the wallet since we have loaded immediately the latest state.

eddycpc (changes requested):

eddycpc:
Screen Shot 2021-06-10 at 10 29 47 AM

We can use this column to customize a little bit the text we show, 
if the tx is of type transfer we can do a few manipulations like 👇

-  If the current wallet address is the one receiving, let's just say here **"Receive NFT"**
-  If the current wallet address is the one sending,  **"Sent NFT"**

If the transaction failed we can just say "Failed Transfer", or "Failed Mint", etc, ...

This way we would be using this one column for many purposes and thus we save space, we also provide much more clarity to the user. We can rename the column from "Type" to "Status", generally most users care about copy text they can understand.

Let me know what you think about this @crypto-matto 🙏

crypto-matto:
Avoid overlapping modals
螢幕截圖 2021-06-10 下午4 13 15
螢幕截圖 2021-06-10 下午4 13 06
螢幕截圖 2021-06-10 下午4 12 59

Optimize NFT Transaction Types
螢幕截圖 2021-06-10 下午4 10 26

PR #521, opened 2021-06-11
Fix QA issues

crypto-matto:
## Fixing QA issues
1. Revise on media previews, showing previews of non-Crypto.com minted NFTs
2. Added more supported video & audio types
3. Bug fix on missing NFTs when GraphQL API crash

### DIFF
--- a/src/service/rpc/ChainIndexingAPI.ts
+++ b/src/service/rpc/ChainIndexingAPI.ts
@@ -70,7 +70,13 @@ export class ChainIndexingAPI implements IChainIndexingAPI {
 
   // eslint-disable-next-line class-methods-use-this
   public async getNftListMarketplaceData(nftLists: NftResponse[]): Promise<NftModel[]> {
+    const returnNftList: NftModel[] = [];
     const payload: MintByCDCRequest[] = nftLists.map(item => {
+      returnNftList[`${item.denomId}-${item.tokenId}`] = {
+        ...item,
+        isMintedByCDC: false,
+        marketplaceLink: '',
+      };
       return {
         denomId: item.denomId,
         tokenIds: [item.tokenId],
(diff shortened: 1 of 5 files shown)


src/service/rpc/ChainIndexingAPI.ts
```diff
@@ -70,7 +70,13 @@ export class ChainIndexingAPI implements IChainIndexingAPI {
 
   // eslint-disable-next-line class-methods-use-this
   public async getNftListMarketplaceData(nftLists: NftResponse[]): Promise<NftModel[]> {
+    const returnNftList: NftModel[] = [];
     const payload: MintByCDCRequest[] = nftLists.map(item => {
+      returnNftList[`${item.denomId}-${item.tokenId}`] = {
+        ...item,
+        isMintedByCDC: false,
+        marketplaceLink: '',
+      };
```
eddycpc, line 79: Can't seem to make sense of what's happening here.

Is `returnNftList` a key-value store ?

crypto-matto: Using `[denomId]-[tokenId]` as the index of the `returnNftList` array, like a key-value store, as denomId & tokenId pair is unique. At the same time, initiate both `isMintedByCDC` and `marketplaceLink` default value. 

Once received the API return, will add back the corresponding `isMintedByCDC` and `marketplaceLink` value to the item in `returnNftList` according to its `[denomId]-[tokenId]` index. So whatever the API returns a full list or not, this function will still return the original full list of `returnNftList`, just that the missing items from the API would include a default `isMintedByCDC` and `marketplaceLink` value.

eddycpc: If it's a key-value store, I wouldn't probably name it a list, that just makes it very hard to reason about it.

I would probably call it `nftListMap` `nftKVStore` or `NftCache`, that gives an idea of what it really is and helps pass the information well to the code reader. When I saw it being called list I expected it to be a linear list data structure and that's why it was confusing to me.

eddycpc (approved):

PR #527, opened 2021-06-21
Reward withdrawal transaction failed. sign fail with Ledger (Nano X) …

leejw51crypto:
pad zeros to 32 bytes for R,S from der format values


leejw51crypto:
some devices return 31 bytes or 33 bytes for r  or s values..
it should be 32 bytes precisely for secp256k1

eddycpc (approved):

PR #531, opened 2021-06-27
Feature - Add NFT mint section

crypto-matto:
螢幕截圖 2021-06-28 上午3 05 28
螢幕截圖 2021-06-28 上午3 04 53
螢幕截圖 2021-06-28 上午3 05 38
螢幕截圖 2021-06-28 上午3 07 23
螢幕截圖 2021-06-28 上午3 07 52

### DIFF
--- a/src/pages/home/home.tsx
+++ b/src/pages/home/home.tsx
@@ -467,17 +467,15 @@ function HomePage() {
     );
   };
 
-  const renderNftTitle = (_nft: NftProcessedModel | undefined, length: number = 0) => {
+  const renderNftTitle = (_nft: NftProcessedModel | undefined, length: number = 999) => {
     if (_nft?.tokenData && _nft?.tokenData.name) {
-      return length > 0 ? ellipsis(_nft?.tokenData.name, length) : _nft?.tokenData.name;
+      return ellipsis(_nft?.tokenData.name, length);
     }
     if (_nft?.tokenData && _nft?.tokenData.drop) {
-      return length > 0 ? ellipsis(_nft?.tokenData.drop, length) : _nft?.tokenData.drop;
+      return ellipsis(_nft?.tokenData.drop, length);
     }
     if (_nft) {
-      return length > 0
-        ? ellipsis(`${_nft?.denomId} - #${_nft?.tokenId}`, length)
-        : `${_nft?.denomId} - #${_nft?.tokenId}`;
+      return ellipsis(`${_nft?.denomId} - #${_nft?.tokenId}`, length);
     }
     return 'n.a.';
   };
--- a/src/pages/home/home.tsx
+++ b/src/pages/home/home.tsx
@@ -846,8 +846,8 @@ function HomePage() {
                     >
                       <Meta
                         title={
-                          item?.tokenData.drop
-                            ? item?.tokenData.drop
+                          item?.tokenData.name
+                            ? item?.tokenData.name
                             : `${item?.denomId} - #${item?.tokenId}`
                         }
                         description={
--- a/src/pages/home/home.tsx
+++ b/src/pages/home/home.tsx
@@ -467,6 +467,21 @@ function HomePage() {
     );
   };
 
+  const renderNftTitle = (_nft: NftProcessedModel | undefined, length: number = 0) => {
+    if (_nft?.tokenData && _nft?.tokenData.name) {
+      return length > 0 ? ellipsis(_nft?.tokenData.name, length) : _nft?.tokenData.name;
+    }
+    if (_nft?.tokenData && _nft?.tokenData.drop) {
+      return length > 0 ? ellipsis(_nft?.tokenData.drop, length) : _nft?.tokenData.drop;
+    }
+    if (_nft) {
+      return length > 0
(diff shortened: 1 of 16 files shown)


src/pages/nft/nft.tsx
```diff
@@ -44,6 +71,619 @@ const { Header, Content, Footer, Sider } = Layout;
...
+        AnalyticsTxType.TransferTransaction,
+        AnalyticsActions.FundsTransfer,
+        AnalyticsCategory.Transfer,
+      );
+
+      // const latestLoadedNFTs = await walletService.retrieveNFTs(currentSession.wallet.identifier);
+      // setNftList(latestLoadedNFTs);
+      // const processedNFTsLists = processNftList(latestLoadedNFTs);
+      // setProcessedNftList(processedNFTsLists);
+
```
eddycpc, line 281: I wonder why this part is commented out, it can be used to sync the latest created NFTs into the local recoil store

src/pages/nft/nft.tsx
```diff
@@ -44,6 +71,619 @@ const { Header, Content, Footer, Sider } = Layout;
...
+    //       resolve()
+    //     }
+    //     waitUntilImageLoaded(resolve, _image);
+    //   }, 1000);
+    // };
+    // await new Promise(resolve => waitUntilImageLoaded(resolve, imageUrl));
+    // const type = 'image/png';
+
+    // setUploading(true);
+
```
eddycpc, line 357: We can remove unused code

src/pages/home/home.tsx
```diff
@@ -846,8 +846,8 @@ function HomePage() {
                     >
                       <Meta
                         title={
-                          item?.tokenData.drop
-                            ? item?.tokenData.drop
+                          item?.tokenData.name
+                            ? item?.tokenData.name
```
eddycpc, line 850: I would suggest we try to use the `drop` value if it's available after realizing the name property does not exist.

So basically I would do something like :
1.  If item `tokenData.name` exists then display it` tokenData.name`
2. If `tokenData.name` does not exist try to show `tokenData.drop`
3. Finally if `tokenData.drop` does not exist then use `${item?.denomId} - #${item?.tokenId}`

eddycpc (changes requested):

src/service/WalletService.ts
```diff
@@ -1133,6 +1134,24 @@ class WalletService {
...
   }
 
+  public async getDenomIdData(denomId: string): Promise<NftDenomModel | null> {
+    const currentSession = await this.storageService.retrieveCurrentSession();
+    if (currentSession?.wallet.config.nodeUrl === NOT_KNOWN_YET_VALUE) {
+      return Promise.resolve(null);
+    }
+    try {
+      const chainIndexAPI = ChainIndexingAPI.init(currentSession.wallet.config.indexingUrl);
+      const nftDenomData = await chainIndexAPI.fetchNftDenomData(denomId);
```
eddycpc, line 1144: I'm curious what do we pass exactly here is it `denomId` or `denom name` ?

That's because in the UI form for mint NFT, the input field is being called **denom name**

eddycpc (changes requested):

src/pages/nft/nft.tsx
```diff
@@ -411,352 +1113,368 @@ const NftPage = () => {
...
+                      </div>
+                    ) : (
+                      ''
+                    )}
+                  </div>
+                </>
+              </ErrorModalPopup>
+            </>
+          </TabPane>
+          <TabPane tab="NFT Mint" key="2">
```
eddycpc, line 1464:
```suggestion
          <TabPane tab="Mint NFT" key="2">
```

src/service/WalletService.ts
crypto-matto: If you look at one of our [official minted NFT](https://crypto.org/explorer/nfts/tokens/k9528183ec3b706ca28d3f6fc5d31cf60/edition63289) , both `denomId` and `denomName` are identical, but Denom Name would be a more user-friendly. 

https://crypto.org/explorer/api/v1/nfts/denoms/k9528183ec3b706ca28d3f6fc5d31cf60

src/pages/home/home.tsx
```diff
@@ -467,6 +467,21 @@ function HomePage() {
     );
   };
 
+  const renderNftTitle = (_nft: NftProcessedModel | undefined, length: number = 0) => {
```
eddycpc, line 470: I think the `length` check in many places is not necessary.

The `ellipsis` function is already doing that in the implementation. Let's remove it and keep the code much cleaner.

You can see in the definition below that the code is doing exactly what you're replicating here.
```
export function ellipsis(str: string, len: number) {
  return str.length <= len ? `${str}` : `${str.substr(0, len)}...`;
}
```

eddycpc (changes requested):

eddycpc (approved):

crypto-matto:
- Card grid image in square-cover
螢幕截圖 2021-06-29 下午5 49 22

- Loading spin when image loading
螢幕截圖 2021-06-29 下午6 28 42

PR #538, opened 2021-07-02
Fix QA issues on NFT Mint

crypto-matto:
- NFT Collection Item per page changes responsively (10 as maximum)
螢幕截圖 2021-07-02 上午10 47 57
螢幕截圖 2021-07-02 上午10 48 06
螢幕截圖 2021-07-02 上午10 48 28

- Optimized Upload flow with additional upload messages:
螢幕截圖 2021-07-02 上午10 25 29
螢幕截圖 2021-07-02 上午10 18 33

- App Crash when Description is empty

- Updated error message on `denomId` & `tokenId`

### DIFF
--- a/src/pages/nft/nft.tsx
+++ b/src/pages/nft/nft.tsx
@@ -38,7 +38,14 @@ import {
   walletAssetState,
   ledgerIsExpertModeState,
 } from '../../recoil/atom';
-import { ellipsis, middleEllipsis, isJson, convertIpfsToHttp, sleep } from '../../utils/utils';
+import {
+  ellipsis,
+  middleEllipsis,
+  isJson,
+  convertIpfsToHttp,
+  sleep,
+  useWindowSize,
+} from '../../utils/utils';
 import { getUINormalScaleAmount } from '../../utils/NumberUtils';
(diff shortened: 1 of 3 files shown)


src/pages/nft/nft.tsx
```diff
@@ -460,7 +487,8 @@ const FormMintNft = () => {
             { required: true, message: 'Denom ID is required' },
             {
               pattern: /(^[a-z](([a-z0-9]){2,63})$)/,
-              message: 'Denom ID can only be alphabetic & between 3 and 64 characters',
+              message:
+                'Denom ID can only start with lowercase alphabetic, followed by lowercase alphabetic or number between 3 and 64 characters',
```
eddycpc, line 491: This error message has become too long, can we break it down into 2 validation objects.

- The first one should check if the value starts with lower case only: `Denom ID can only start with lowercase alphabetic`

- The second error message should finally check the second part of the error message: `followed by lowercase alphabetic or number between 3 and 64 characters`

This way we will be reducing drastically the length of the error message to something short and more readable 🤔

eddycpc (changes requested):

eddycpc (approved):

PR #579, opened 2021-07-15
[Phrase Export] (Fix: #578) Add capability to export phrase

crypto-matto:
Added export button in `Settings` => `General Configuration`
螢幕截圖 2021-07-15 下午8 42 36

Password entry required
螢幕截圖 2021-07-15 下午8 32 57

UIUX Referencing DeFi Wallet:
The choice of 3 items per page is the highest common factor of all possible total number of words in recovery phrase: `12,15,18,21,24`
螢幕截圖 2021-07-15 下午8 28 56

Image from iOS

### DIFF
--- a/src/pages/settings/settings.tsx
+++ b/src/pages/settings/settings.tsx
@@ -380,7 +380,7 @@ function MetaInfoComponent() {
             <div className="title">Data Analytics</div>
             <div className="description">
               The data collected for analytics is used to prioritize development for new features
-              and functionalities and also to improve implemented features
+              and functionalities and also to improve implemented features.
             </div>
             <Switch
               checked={!defaultGAStateDisabled}
@@ -391,7 +391,7 @@ function MetaInfoComponent() {
           </div>
           <Divider />
           <div className="item">
-            <div className="title">Export your recovery phrase</div>
+            <div className="title">Export your Recovery Phrase</div>
             <div className="description">
               You are required to enter your App Password for this action.
             </div>
--- a/src/pages/settings/settings.tsx
+++ b/src/pages/settings/settings.tsx
@@ -197,11 +197,13 @@ function MetaInfoComponent() {
   const [isExportRecoveryPhraseModalVisible, setIsExportRecoveryPhraseModalVisible] = useState<
     boolean
   >(false);
+  const [currentSlide, setCurrentSlide] = useState(0);
 
   const didMountRef = useRef(false);
 
   // Slider related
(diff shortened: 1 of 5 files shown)


src/pages/settings/settings.tsx
```diff
@@ -290,8 +389,134 @@ function MetaInfoComponent() {
...
+      <ModalPopup
+        className="export-recovery-phrase-modal"
+        isModalVisible={isExportRecoveryPhraseModalVisible}
+        handleCancel={() => {
+          setIsExportRecoveryPhraseModalVisible(false);
+          recoveryPhraseSlider?.goTo(0);
+        }}
+        handleOk={() => setIsExportRecoveryPhraseModalVisible(false)}
+        // confirmationLoading={isButtonLoading}
+        // closable={!isButtonLoading}
```
eddycpc, line 439: Let's clean unused code

src/pages/settings/settings.tsx
```diff
@@ -290,8 +389,134 @@ function MetaInfoComponent() {
...
+            >
+              Copy
+            </Button>
+          </CopyToClipboard>,
+          <Button
+            key="back"
+            type="link"
+            onClick={() => {
+              setIsExportRecoveryPhraseModalVisible(false);
+              recoveryPhraseSlider?.goTo(0);
```
eddycpc, line 456: We can also clear the decryptedPhrase on close that way it doesn't stay for long in the memory.

We can set its value to `null` or `''` on close

eddycpc (changes requested):

eddycpc (commented):
Superb work on the UI !, overall looks great, I left some comments

src/pages/settings/settings.tsx
```diff
@@ -290,8 +389,134 @@ function MetaInfoComponent() {
...
+            <div className="title">Export your Recovery Phrase</div>
+            <div className="description">
+              You are required to enter your App Password for this action.
+            </div>
+            <Button
+              type="primary"
+              onClick={() => {
+                showPasswordInput();
+              }}
+            >
```
eddycpc, line 403: One thing we can do here is to hide this section if the wallet type is LEDGER, that's because LEDGER wallet types don't have the phrase stored in the app.

eddycpc (changes requested):

src/pages/settings/settings.tsx
crypto-matto: Oh yes indeed, I forgot this one.

eddycpc (approved):

Issue #575, opened 2021-07-12
Languages and Internationalization support on the app


crypto-3dd:
- [x] Add UI settings to select language
- [x] Integrate with a multi-language supported library
- [x] Add strings for internationalization and replace all static in-app strings
- [x] Pick a UI tool or even spreadsheets to hold translations and write a script to download the translation into the format to be used directly in the app.

crypto-matto:
Google Spreadsheet for multi-language management:
https://docs.google.com/spreadsheets/d/1QTqPb97jr6zOD7krA4pAXHYx51CiSkJYq7m71OYcLok/edit#gid=0

App Script
https://script.google.com/home/projects/1X1WCMV0tbQsXvlkV70Kw_3gvF85zSxeqkUnTuDhul_EcdWG9ZU4_Z6ow/edit

crypto-matto:
I'll make sure we would have at least 2 languages support first, then we may merge this PR & role out this feature.

PR #583, opened 2021-07-21
[Feature] (Fix: #575) Languages and Internationalization support on the app

crypto-matto:


src/language/I18n.ts
```diff
@@ -0,0 +1,35 @@
...
+  // detect user language
+  // learn more: https://github.com/i18next/i18next-browser-languageDetector
+  .use(LanguageDetector)
+  // pass the i18n instance to react-i18next.
+  .use(initReactI18next)
+  // init i18next
+  // for all options read: https://www.i18next.com/overview/configuration-options
+  .init({
+    debug: true,
+    lng: savedLanguage,
```
eddycpc, line 27:
```suggestion
    lng: await generalConfigService.getLanguage(),
```

I wonder if it's possible to call it directly this way and thus avoiding the init method `getLanguageFromDB` altogether

src/models/Wallet.ts
```diff
@@ -57,6 +57,11 @@ export interface SettingsDataUpdate {
   networkFee?: string | undefined;
 }
 
+export interface LanguageSettings {
+  walletId: string;
+  langaugeCode: string;
+}
+
```
eddycpc, line 64: Since LanguageSettings settings go beyond a single wallet, it's an app-wide configuration, can you add it in the `GeneralConfig` model directly?

src/language/I18n.ts
crypto-matto: While simply using the suggested change is not possible, as `i18n.init()` is not an `async` funciton, but I've managed to use a proper implementation with using `LanguageDetector` Instance. 
```
const languageDetector = new LanguageDetector(async () => {
  return await generalConfigService.getLanguage();
});

i18n
  // detect user language
  // learn more: https://github.com/i18next/i18next-browser-languageDetector
  .use(languageDetector)
  // pass the i18n instance to react-i18next.
  .use(initReactI18next)
  // init i18next
  // for all options read: https://www.i18next.com/overview/configuration-options
  .init({
    debug: true,
    fallbackLng: 'enUS',
    interpolation: {
      escapeValue: false, // not needed for react as it escapes by default
    },
    resources: translations,
  });

export default i18n;
```

src/models/Wallet.ts
crypto-matto: Oh this one has become obsoleted, just forgot to delete. 
I'm already putting `languageCode` inside `GeneralConfig`
```
export interface GeneralConfig {
  hasEverShownAnalyticsPopup: boolean;
  languageCode: string;
}
```

src/language/code.gs
```diff
@@ -0,0 +1,336 @@
+/*
```
eddycpc, line 1: Can we call this script `csv to json transformer` or something more descriptive. The first time I read the file 'code.gs', I couldn't know what it's supposed to do.

Let's call it something more descriptive.

eddycpc (commented):
Overall looks good.

An important addition you can do is to add in the README file, a section explaining what to script(s) to run in order to download the latest version of translation strings.

crypto-3dd (approved):
Looks good to me!

We can go ahead and invite translators to update the translations strings

crypto-3dd:
@crypto-matto You might have to fix merge conflicts

crypto-3dd:
> @crypto-matto You might have to fix merge conflicts

Never mind, I've fixed it now.

crypto-3dd (approved):

crypto-3dd:
We shall merge this for now, and keep iterating on adding languages when they're complete since it's now released

crypto-3dd (approved):

Issue #586, opened 2021-07-22
Use fixed version on dependencies


crypto-3dd:

crypto-3dd:
Fixed in https://github.com/crypto-com/chain-desktop-wallet/pull/587

PR #587, opened 2021-07-22
Fix and lock dependencies

crypto-3dd:
Related to #586


crypto-matto (dismissed):

crypto-matto:
Make sure to commit the new `yarn.lock` file too?

Issue #598, opened 2021-07-27
Create a Cronos Rpc client service that interacts with the node for balance fetching, transaction broadcast, etc, ....


crypto-3dd:
Create a Cronos Rpc client that interacts with the node for balance fetching, transaction broadcast, transaction history fetching.

This RPC client service/class + interface should simplify the interactions between the wallet and the ethereum-like nodes.
Roles :

- [x] Fetching address balances.
- [x] Fetching address transaction history, the previous 20 transactions or more.
- [x] Broadcast transactions on-chain.
- [x] Should be flexible to work out of the box for all Ethereum-Like assets that would be supported later.

cdc-Hitesh:
Do we want to:
- directly consume Web3 Provider in our Chain desktop (eg, infura/alchemy provider) or
- create a backend API service with ethereum related functionality (using nestjs etc)?

crypto-3dd:
Just to clarify we will not have to create a new backend, I don't think it will be necessary.
As the ticket describes it, all we need is a simple class+interface that knows how to EVM broadcast transactions, fetch balances and load the latest transactions given an account address.

```javascript
interface EVMRpC {

function loadBalance(address: string) {}

function broadcastTransaction(signedTxHex: string) {}

function loadTransactions(address: string) {}

function fetchNonce(address: string) {}

//// etc

}
```

cdc-Hitesh:
Noted. Thanks for the clarification.

cdc-Hitesh:
Closed the issue

PR #613, opened 2021-08-03
#598: Introduced RPC and ChainIndexing Interfaces

cdc-Hitesh:
- [x] interface
- [x] service business logic
- [x] EvmClient
- [x] CronosClient
- [ ] CronosClient - ChainIndexing API business logic


.github/workflows/build.yml
```diff
@@ -7,8 +7,6 @@ on:
 jobs:
   release:
     runs-on: ${{ matrix.os }}
-    env:
-      NODE_OPTIONS: '--max_old_space_size=4096'
```
crypto-3dd, line 11: I wonder if there is a reason you are modifying the build actions' latest updates? I would suggest to possibly bring back to the normal state from `cronos-general-support`

crypto-3dd (changes requested):

package.json
```diff
@@ -41,7 +41,6 @@
     "electron-log": "4.3.5",
     "electron-reload": "1.5.0",
     "electron-updater": "4.3.9",
-    "ethers": "^5.4.2",
```
crypto-3dd, line 44: Did you remove `ethers` dependency?

src/config/StaticAssets.ts
```diff
@@ -1,40 +0,0 @@
...
-      'https://firebasestorage.googleapis.com/v0/b/chain-desktop-wallet.appspot.com/o/cronos_logo.png?alt=media&token=781c48a3-e89e-4dd4-87d3-d1a1b8e2e456',
-    identifier: getRandomId(),
-    name: 'Cronos Coin',
-    symbol: 'CRONOS',
-    mainnetSymbol: 'CRONOS', // This is to be used solely for markets data since testnet market prices is always non existent
-    stakedBalance: '0',
-    decimals: 18,
-    assetType: UserAssetType.EVM,
-  };
-};
```
crypto-3dd, line 40: I think there have been some issues when you were merging from `cronos-general-support`

src/config/StaticConfig.ts
```diff
@@ -193,6 +193,24 @@ export const DefaultWalletConfigs = {
...
+      'https://s3-ap-southeast-1.amazonaws.com/monaco-cointrack-production/uploads/coin/colorful_logo/5c1248c15568a4017c20aa87/cro.png',
+    identifier: getRandomId(),
+    name: 'Crypto.org Coin',
+    symbol: assetSymbol,
+    mainnetSymbol: 'CRO', // This is to be used solely for markets data since testnet market prices is always non existent
+    stakedBalance: '0',
+    decimals: 8,
+  };
+};
+
```
crypto-3dd, line 213: Normally  this shouldn't be added here again

package.json
cdc-Hitesh: These changes are somehow picked up after reverting my eslint based commits.

.github/workflows/build.yml
cdc-Hitesh: These changes are somehow picked up after reverting my eslint based commits.

src/config/StaticAssets.ts
cdc-Hitesh: I agree. Something is wrong with the merging and reverting.

src/config/StaticConfig.ts
cdc-Hitesh: I agree something is wrong while reverting and merging.

PR #617, opened 2021-08-04
Cronos Backend Support  -  Step 2

crypto-3dd:
- [x] Implement EVM Transaction Signer service and add unit tests
- [x] Add asset configuration: Things like support for Ledger, Staking, etc, should be disabled initially on CRONOS
- [x] Update balance fetching for multi-assets support - Test
- [ ] Dynamic asset price fetching for cronos and other EVM assets.
- [ ] Update transactions fetching for multi-assets support.
- [ ] Add a flow to request the user the password to re-create missing assets since the phrase is required for assets creation.


src/service/signers/EvmTransactionSigner.ts
```diff
@@ -1 +1,40 @@
...
+import {
+  DelegateTransactionUnsigned,
+  TransferTransactionUnsigned,
+  WithdrawStakingRewardUnsigned,
+} from './TransactionSupported';
+
+export class EvmTransactionSigner implements ITransactionSigner {
+  // eslint-disable-next-line class-methods-use-this
+  public signTransfer(transaction: TransferTransactionUnsigned, phrase: string): Promise<string> {
+    const txParams = {
```
cdc-Hitesh, line 13: shall we introduce 
- address checksum checks ?
- default `gasLimit` value ?

src/service/signers/EvmTransactionSigner.ts
```diff
@@ -1 +1,40 @@
...
+      data: '0x',
+    };
+
+    const tx = new Transaction(txParams);
+
+    const { privateKey } = ethers.Wallet.fromMnemonic(phrase);
+    tx.sign(Buffer.from(privateKey.replace('0x', ''), 'hex'));
+
+    const signedTransactionHex = `0x${tx.serialize().toString('hex')}`;
+    return Promise.resolve(signedTransactionHex);
```
cdc-Hitesh, line 28: can we introduce a hex validation check for the final `signedTransactionHex`?

Though it may be an overkill but as a thought.

crypto-3dd: Since this is not a user input provide data, that's probably unnecessary to do any sort of validation at this level.

src/service/signers/EvmTransactionSigner.ts
crypto-3dd: - At the signer level, his job is just to sign transactions. ( **single responsibility principle** )
- Checksum validations would happen where the recipient address is being input.
- The signer is not required to know about default `gasLimit` since that's the job of the wallet service.

Just to recap, the signer's job is to get some param data and sign transactions, this part of the code shouldn't know too much.

PR #620, opened 2021-08-04
Problem: Add EVM and CronosClient

cdc-Hitesh:
- [x] EVM Interface
- [x] Cronos Chain Index API Interface
- [x] Cronos Client ( uses chainIndexAPI and EVMClient)
- [x] Request / Response Models
- [x] Business logic for EVMClient and CronosClient
- [x] Unit tests EVMClient using nock
- [x] Unit tests CronosClient for chain indexing APIs


src/service/rpc/clients/EVMClient.ts
```diff
@@ -0,0 +1,103 @@
...
+
+
+    // Node
+    async isNodeSyncing(): Promise<boolean> {
+        const isSyncing = await this.web3.eth.isSyncing();
+        if (isSyncing instanceof Object) {
+            return true;
+        }
+        return false;
+    }
```
crypto-3dd, line 30: I don't think this is really needed, especially on the desktop wallet, let me know if there's a use case for this one currently

crypto-3dd (commented):
This is looking pretty good already, good start!

src/service/cronos/CronosClient.ts
```diff
@@ -0,0 +1,24 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
+import { EVMClient } from "../rpc/clients/EVMClient";
+import { ICronosChainIndexAPI } from "../rpc/interface/cronos.chainIndex";
+import { TxListAPIResponse } from "../rpc/models/cronos.models";
+
```
crypto-3dd, line 5: Overall, all looks good, I would suggest now also to add some mock tests, these tests help document how to use the API.

Have a look at how we are doing in the `NodeRpcService.spec` test class

src/service/rpc/clients/EVMClient.ts
cdc-Hitesh: I have observed whenever we are working with:
- transaction receipt fetching 
- Showing users the status of our connected nodes

checking the node syncing status helps.

Still, let me know if you want it to be removed.

src/service/cronos/CronosClient.ts
cdc-Hitesh: Noted.
I will add.

src/service/cronos/CronosClient.ts
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * name: CronosClient
+ * purpose: This client can be used to handle `Cronos` related operations.
+ */
+
+class CronosClient extends EVMClient implements ICronosChainIndexAPI {
+
+    getTxsByAddress = async (_address: string): Promise<TxListAPIResponse> => {
+        // todo: implement the logic using `axios` client
+        throw new Error("Method not implemented.");
```
crypto-3dd, line 15: You haven't found yet a service to pull transactions history data?

You might ask Calvin or any other core dev working on `Cronos`, they might help on this

crypto-3dd (changes requested):

src/service/cronos/CronosClient.ts
cdc-Hitesh: Hey @crypto-3dd 
I didn't push it sorry. It was done locally.
Thanks

src/service/cronos/CronosClient.ts
```diff
@@ -0,0 +1,71 @@
...
+    txPendingListRequestOptions,
+} from '../rpc/interface/cronos.chainIndex';
+import {
+    TxListAPIResponse,
+    txListByAccountRequestParams,
+    PendingTxListAPIResponse,
+} from '../rpc/models/cronos.models';
+import axios, { AxiosResponse } from 'axios';
+
+const cronosExplorerAPIBaseURI = 'https://cronos-explorer.crypto.org/api';
```
crypto-3dd, line 15: Instead of having this value hardcoded, I would suggest having it as a constructor property, that would add more flexibility for the `CronosClient` class. And the explorer URL can be UI configurable object.

crypto-3dd (changes requested):

src/service/rpc/clients/EVMClient.ts
```diff
@@ -0,0 +1,110 @@
...
+        throw new Error("Please provide a valid HTTP Web3 Provider.");
+    }
+
+    get web3(): Web3 {
+        return this._web3;
+    }
+
+    set web3(web3: Web3) {
+        this._web3 = web3;
+    }
```
crypto-3dd, line 27: I wonder if this `setter` is needed if the web3 param can be set immediately from the constructor?

I would probably make the web3 immutable and just keep the constructor, the getter and setter would not be needed doing that way.

src/service/cronos/CronosClient.ts
cdc-Hitesh: Okay sure it makes sense.
Also, I think this value might come from any `StaticConfig` variable.

src/service/rpc/clients/EVMClient.ts
cdc-Hitesh: I have just kept the `setter` for future-ready scenario. Let's say if the web3 instance is to be changed on the fly or creating a fallback mechanism.

But again, there is no concrete future use case I see.

Do you want me to remove it or keep it for now?

crypto-matto:
The `yarn audit` error shall be fixed by adding an updated enough package version, as suggested in the error, under `package.json` => `resolutions`
```
  "resolutions": {
    ...,
    "tar": "6.1.1"
  }
```

cdc-Hitesh:
> The `yarn audit` error shall be fixed by adding an updated enough package version, as suggested in the error, under `package.json` => `resolutions`
> 
> ```
>   "resolutions": {
>     ...,
>     "tar": "6.1.1"
>   }
> ```

Thanks Matthew.

crypto-3dd (approved):

PR #621, opened 2021-08-09
Clean inappropriate code

crypto-matto:


src/pages/staking/staking.tsx
```diff
@@ -595,8 +589,8 @@ const FormWithdrawStakingReward = () => {
   // TO-DO temporary CRO Asset select
   const assetType =
     currentSession.wallet.config.name === DefaultWalletConfigs.TestNetConfig.name
-      ? UserAssetType.TENDERMINT
-      : UserAssetType.TENDERMINT;
+      ? DefaultWalletConfigs.TestNetConfig.assetIdentifer?.cro
+      : DefaultWalletConfigs.MainNetConfig.assetIdentifer?.cro;
   const croAsset = TransactionUtils.getAssetFromAllAssets(walletAllAssets, assetType);
   setWalletAsset(croAsset);
```
crypto-3dd, line 595: This is still not very scalable, you will need to remove entirely `TransactionUtils.getAssetFromAllAssets`, it's not a good pattern if we want to scale to multiple assets.

src/config/StaticConfig.ts
```diff
@@ -46,6 +46,10 @@ export type WalletConfig = {
     gasLimit: string;
     networkFee: string;
   };
+  assetIdentifer?: {
+    cro: string;
+    cronos: string;
+  };
```
crypto-3dd, line 52: This doesn't scale very well as assets count, we can't hard code `assetID`, would you add this for every asset we add in the future.

I still don't know why we would need to hard code this, can you please elaborate.

src/config/StaticConfig.ts
```diff
@@ -156,6 +160,10 @@ const MainNetConfig: WalletConfig = {
     gasLimit: FIXED_DEFAULT_GAS_LIMIT,
     networkFee: FIXED_DEFAULT_FEE,
   },
+  assetIdentifer: {
+    cro: 'e9cc453a9e39718c',
+    cronos: 'd1a17aca33fd382c',
+  },
```
crypto-3dd, line 166: We can't hardcode these values, they are created on the fly and shouldn't checked against.

crypto-matto: I see. I thought the id of cro & cronos are always the same. I misunderstood it.

src/pages/staking/staking.tsx
crypto-matto: Right now CRONOS is set as the only non-secondary asset. Yet I still have to pick CRO from allAsset to support CRO-only delegation related operations for 1st release. That’s why I need to use a function like `getAssetFromAllAssets()` (Temporary) to get me CRO from allAsset , and I used `identifier` & `assetType` (which I misunderstood its uniqueness).

PR #642, opened 2021-08-24
[EVM] Cronos Frontend Step 6

crypto-matto:
- [x] Move NFT Transactions from `Home` to `My NFT`
 - [x] Fixed Wallet List Menu Key
 - [x] Fixed Transaction records incorrect display issues
 - [x] Removed `send.tsx` and Sender Address error
 - [x] Implement part of updated feedback from Design Team (hidden at the moment)

螢幕截圖 2021-08-25 下午5 59 21
螢幕截圖 2021-08-25 下午6 00 08


src/pages/assets/assets.tsx
```diff
@@ -347,6 +357,44 @@ const AssetsPage = () => {
...
+                    //           </>
+                    //           : <>
+                    //             <Icon className="tab" component={IconReceive} />
+                    //             {tab.props.tab}
+                    //           </>}
+                    //       </div>
+                    //     })} */}
+                    //     </div>
+                    //   );
+                    // }}
```
crypto-3dd, line 397: Let's remove unused stuff

src/pages/assets/components/FormSend.tsx
```diff
@@ -281,7 +281,7 @@ const FormSend: React.FC<FormSendProps> = props => {
             <div className="description">{t('send.modal1.description')}</div>
             <div className="item">
               <div className="label">{t('send.modal1.label1')}</div>
-              <div className="address">{`${currentSession.wallet.address}`}</div>
+              <div className="address">{`${currentSession.activeAsset?.address}`}</div>
```
crypto-3dd, line 284:
```suggestion
              <div className="address">{`${currentSession.activeAsset?.address || currentSession.wallet.address }`}</div>
```
For backward compatibility.

src/pages/assets/assets.tsx
crypto-matto: This is some of the updated design from Design Team. But I hide it for a moment since it's not settled.

src/pages/home/home.tsx
```diff
@@ -649,26 +321,14 @@ const HomePage = () => {
...
-      const nftTransferTabularData = convertNftTransfers(allNftTransfer);
-
-      showWalletStateNotification(sessionData.wallet.config);
-      checkNewlyAddedStaticAssets(sessionData);
-
-      // setTransfers(transferTabularData);
-      setNftTransfers(nftTransferTabularData);
-      // setUserAsset(currentAsset);
+      showWalletStateNotification(currentSession.wallet.config);
+      checkNewlyAddedStaticAssets(currentSession);
```
crypto-3dd, line 331:
```suggestion
      showWalletStateNotification(sessionData.wallet.config);
      checkNewlyAddedStaticAssets(sessionData);
```

Inside `useEffect` here, `currentSession` is not set yet, it's better to use `sessionData` which is guaranteed to have been set.

crypto-3dd (changes requested):

crypto-matto:
Unhidden the TBC Design
螢幕截圖 2021-08-25 下午6 18 52

crypto-3dd:
- Looking at the new designs for asset detail, how do we get back to the transaction list view ?
- We might need a third icon for transaction listing, which can be clicked and it will then show transactions.
- Minor suggestion, we can add some transition effect if it's doable when changing views, to smooth out the changes from one view to another.

crypto-3dd (approved):

PR #647, opened 2021-08-25
Problem #636: Base denom field added

### LINKED ISSUE #636: Suggestion: Add unit to network fee configuration
Problem: Right now the network fee is an input field with no unit display. It may be confusing to new users. I suggest to add the unit `basecro` or `basetcro` to make it clear

Screenshot 2021-08-16 at 3 29 10 PM

cdc-Hitesh:
Fixes #636 
Added a readonly field in the settings page.

image

### DIFF
--- a/src/language/translations.json
+++ b/src/language/translations.json
@@ -749,9 +749,6 @@
           "gasLimit": {
             "label": "Gas Limit"
           },
-          "denomName": {
-            "label": "Base Denom"
-          },
           "checkbox1": {
             "description1": "Propagate the settings changes to all your ",
             "description2": "wallets"
@@ -1561,9 +1558,6 @@
           "gasLimit": {
             "label": "Gas 費用"
           },
-          "denomName": {
-            "label": "基本面值"
-          },
           "checkbox1": {
             "description1": "將設定更新至所有",
             "description2": "錢包"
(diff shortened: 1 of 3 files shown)


crypto-3dd:
@crypto-matto  @cdc-Hitesh 

I think the issue here is more related to what is the best way to display fees ( gas limit and gas price ) related to each asset.

An example :

-  ETH/EVM gas limits are usually a direct number without a symbol (gas): 21 000, 60 000, etc, ... But the gas price is usually in gwei. 80 `gwei`  >> https://etherscan.io/gastracker.

- CRO uses the `basecro` denomination to the network fee (gas fee), so the solution, we should come with should take into account these realities

cdc-Hitesh:
> @crypto-matto @cdc-Hitesh
> 
> I think the issue here is more related to what is the best way to display fees ( gas limit and gas price ) related to each asset.
> 
> An example :
> 
> * ETH/EVM gas limits are usually a direct number without a symbol (gas): 21 000, 60 000, etc, ... But the gas price is usually in gwei. 80 `gwei`  >> https://etherscan.io/gastracker.
> * CRO uses the `basecro` denomination to the network fee (gas fee), so the solution, we should come with should take into account these realities

Thanks for the note.
I agree, basically there is a need to display network fee `baseUnit` .
WIll update if anything from my side.

crypto-3dd:
Let's re-open when we figure out a clean workable approach

PR #654, opened 2021-08-30
Code clean up - Proper usage and handling of fees on EVM transactions

crypto-3dd:
- Proper handling of gasLimit and gasFee for cronos transactions
- Clean up code, use proper conversions utility methods

### DIFF
--- a/src/service/WalletService.ts
+++ b/src/service/WalletService.ts
@@ -1,5 +1,7 @@
 import axios from 'axios';
 import _ from 'lodash';
+import { TransactionConfig } from 'web3-eth';
+import Web3 from 'web3';
 import {
   DisableDefaultMemoSettings,
   DisableGASettings,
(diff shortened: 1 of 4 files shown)


src/service/WalletService.ts
```diff
@@ -120,12 +122,32 @@ class WalletService {
...
 
           transfer.nonce = await cronosClient.getNextNonceByAddress(currentAsset.address);
 
+          const loadedGasPrice = web3.utils.toWei(
+            await cronosClient.getEstimatedGasPrice(),
+            'gwei',
+          );
+          transfer.gasPrice = Number(loadedGasPrice);
+
+          transfer.gasLimit = Number(await cronosClient.estimateGas(txConfig));
```
mannutech, line 143: This `Number` typecasting may not be safe here.

We can switch to `big.js`

mannutech (changes requested):

src/service/WalletService.ts
crypto-3dd: > This Number typecasting may not be safe here.

Can you please elaborate on how unsafe it can be here?

Let me provide more details : 
- The gas estimate value here is guaranteed to forever fit in Number type and it doesn't have floating-point properties. it's a whole integer number.
- The price estimate is being returned in Gwei, which is guaranteed also to fit inside the Number type. The conversion to the small unit (Wei) is happening inside the `EvmTransactionSigner` class.

I might possibly miss something, let me know if there's something I'm missing on the unsafe part. Thanks.

cdc-Hitesh: Thanks for the explanation @crypto-3dd 

Generally, within blockchain value calculations for numbers, we use any big number handling library. Hence brought that general suggestion here.

Your explanation kind of clarifies the doubt.

cdc-Hitesh: I re-visited, looks good to me resolving.

cdc-Hitesh (approved):

PR #655, opened 2021-08-31
[EVM] Cronos Frontend Step9

crypto-matto:
- [x] Add `Currency` settings
- [x] Display market data with selected `Currency`
- [x] Merge `Send` & `Receive` Page into `Assets` Page


src/config/StaticConfig.ts
```diff
@@ -36,6 +49,7 @@ export type WalletConfig = {
   // E.g: User updates nodeURL in one mainnet wallet, all other mainnet wallets will have the new nodeURL
   enableGeneralSettings: boolean;
   analyticsDisabled: boolean;
+  defaultCurrency: string;
```
crypto-3dd, line 52: Hey @crypto-matto I wonder if we really need this new property `defaultCurrency`  

- We already have `session.currency` which holds USD as the default currency, so we can just asssume whatever is stored inside `session.currency` is the current currency. 

- And when it's updated we would just update the session object and set the new currency.

- So we might not need an extra variable to track the fiat currency currently selected. That keeps much simpler since we are only tracking a single state variable.

crypto-3dd (changes requested):

src/config/StaticConfig.ts
crypto-matto: Hi @crypto-3dd, sorry for causing the confusion. I forgot to delete this field here since I was using `defaultCurrency` in earlier commit, but I switched to use `session.currency` in commit https://github.com/crypto-com/chain-desktop-wallet/pull/655/commits/27bcd2a3f2f19b4cae77106f247aeae1039f9dbd .

src/pages/settings/settings.tsx
```diff
@@ -355,6 +368,34 @@ function MetaInfoComponent() {
...
+    setSession(newSession);
+    await walletService.loadAndSaveAssetPrices(newSession);
+
+    const userAsset = walletAllAssets.filter(asset => {
+      return !asset.isSecondaryAsset;
+    });
+    const currentMarketData = await walletService.retrieveAssetPrice(
+      userAsset[0]?.mainnetSymbol,
+      newSession.currency,
+    );
```
crypto-3dd, line 392: This section might not scale well when new currencies are added.

You seem to hardcode currency data to only the primary asset, this section should work for all assets. 
If we want to hold market data in the setMarketData, it should be asset agnostic, it means when new assets are added it should normally work nevertheless.

crypto-3dd (changes requested):

src/pages/settings/settings.tsx
crypto-matto: Not sure if you want me to hardcode 'CRO' as the input parameter, or use either:
1. `walletService.retrieveDefaultWalletAsset()` (which is basically running the same for loop logic & pick the first asset)
2. `session.activeAsset`

I'm hardcoding to 'CRO' for the moment.

crypto-3dd: What I meant is the market data is needed for all assets in the wallet, not only the primary asset.

If a user has 4 assets in the wallet they would want to have all assets with fiat currency conversion, not only the primary one.

src/config/StaticConfig.ts
```diff
@@ -24,6 +24,20 @@ export const SUPPORTED_LANGUAGE = [
   { value: 'koKR', label: '한국어' },
 ];
 
+export const DEFAULT_ASSET_SYMBOL = 'CRO';
```
crypto-3dd, line 27: Sorry, this is not very flexible. I don't understand yet why we would need to hardcode this.

crypto-3dd (changes requested):

src/config/StaticConfig.ts
```diff
@@ -24,6 +24,20 @@ export const SUPPORTED_LANGUAGE = [
...
+export const DEFAULT_CURRENCY_CODE = 'USD';
```
crypto-3dd, line 28: This is also not needed because the default fiat currency code is already held in the `session.currency` property

src/pages/settings/settings.tsx
```diff
@@ -249,10 +257,12 @@ const GeneralSettingsForm = props => {
 
 function MetaInfoComponent() {
   const [session, setSession] = useRecoilState(sessionState);
+  const setMarketData = useSetRecoilState(marketState);
   const [updateLoading, setUpdateLoading] = useState(false);
   const { walletType } = session.wallet;
 
   const [defaultLanguageState, setDefaultLanguageState] = useState<string>(DEFAULT_LANGUAGE_CODE);
+  const [defaultCurrencyState, setDefaultCurrencyState] = useState<string>(DEFAULT_CURRENCY_CODE);
```
crypto-3dd, line 265:
```suggestion
  const [defaultCurrencyState, setDefaultCurrencyState] = useState<string>(currentSession.currency);
```

src/pages/settings/settings.tsx
```diff
@@ -355,6 +368,32 @@ function MetaInfoComponent() {
...
+    const newSession = {
+      ...session,
+      currency: value.toString(),
+    };
+    await walletService.setCurrentSession(newSession);
+    setSession(newSession);
+    await walletService.loadAndSaveAssetPrices(newSession);
+
+    const currentMarketData = await walletService.retrieveAssetPrice(
+      DEFAULT_ASSET_SYMBOL,
```
crypto-3dd, line 387: This can be problematic, it only fetches fiat currency data for **only one asset**.

If a wallet has more than 1 asset, **it's good to go ahead and fetch currency conversions for all the assets** in the wallet, not only one asset.

The proper approach here is to fetch `currencyMarketData` for all assets in the wallet

crypto-3dd (changes requested):

crypto-3dd:
Screen Shot 2021-09-01 at 9 56 51 AM

British Pound having a $ sign in front. I would we can make it flexible so that in the `SUPPORTED_CURRENCY`, we specify a symbol.

That way it can be picked up easily. Something close to the code below.

`SUPPORTED_CURRENCY[currentSession.currency].getSymbol()`

At the same time, `SUPPORTED_CURRENCY` can be given a type as an array or a map of Currencies i.e. `CurrencySupported`

```
export interface CurrencySupported {
    value: string;
    symbol: string;
    label: string;
}
```

src/pages/settings/settings.tsx
crypto-matto: Got your idea. I'll update in other places as well.

src/storage/StorageService.ts
```diff
@@ -228,6 +228,10 @@ export class StorageService {
     );
   }
 
+  public async retrieveAllAssetsPrices(currency: string) {
+    return this.db.marketPriceStore.find<AssetMarketPrice>({ _id: new RegExp(currency) });
```
crypto-3dd, line 232: How does this `_id: new RegExp(currency)` work?

src/config/StaticConfig.ts
```diff
@@ -24,6 +24,18 @@ export const SUPPORTED_LANGUAGE = [
   { value: 'koKR', label: '한국어' },
 ];
 
+export const SUPPORTED_CURRENCY = [
+  { value: 'USD', label: 'USD - $', symbol: '$' },
```
crypto-3dd, line 28: You can make this a typed array as suggested below, it would make it hard to mess up accidentally. And strong types are always preferred when possible.

src/storage/StorageService.ts
crypto-matto: The `_id` are stored as:
- `CRO-USD`
- `CRO-GBP`
- `CRO-EUR`
- `CRONOS-USD`
- `ATOM-USD`
...
Therefore, buy using `new RegExp(currency)` it shall return all `_id`s containing the `currency` parameter.

crypto-3dd: That's very smart I didn't know NeDB allowed those types of queries.

crypto-3dd (approved):

PR #658, opened 2021-08-31
Moved balance fetching to cosmos client

### LINKED ISSUE #657: Problem : Move all obsolete tendermint client calls to the new cosmos client - Croeseid4 compatibility

crypto-3dd:
Fixes #657

### DIFF
--- a/src/service/rpc/NodeRpcService.ts
+++ b/src/service/rpc/NodeRpcService.ts
@@ -5,6 +5,7 @@ import { Bytes } from '../../utils/ChainJsLib';
 import { NodePorts } from '../../config/StaticConfig';
 import {
   AllProposalResponse,
+  BalanceResponse,
   DelegationResult,
   FinalTallyResult,
   LoadedTallyResponse,
@@ -64,8 +65,11 @@ export class NodeRpcService implements INodeRpcService {
   }
 
   public async loadAccountBalance(address: string, assetDenom: string): Promise<string> {
-    const response = await this.tendermintClient.getBalance(address, assetDenom);
-    return response?.amount ?? '0';
+    const response = await this.cosmosClient.get<BalanceResponse>(
+      `/cosmos/bank/v1beta1/balances/${address}/${assetDenom}`,
+    );
+    const balanceData = response?.data;
+    return balanceData?.balance?.amount ?? '0';
   }
 
   public async loadSequenceNumber(address: string): Promise<number> {
--- a/src/service/WalletService.ts
+++ b/src/service/WalletService.ts
@@ -679,6 +679,10 @@ class WalletService {
             currentSession.wallet.address,
             baseDenomination,
           );
+
+          // eslint-disable-next-line no-console
+          console.log('asset.balance.ddz', currentSession.wallet.address, asset.balance);
+
           asset.stakedBalance = await nodeRpc.loadStakingBalance(
             currentSession.wallet.address,
             baseDenomination,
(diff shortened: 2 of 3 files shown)


src/service/rpc/NodeRpcService.ts
```diff
@@ -64,8 +65,11 @@ export class NodeRpcService implements INodeRpcService {
   }
 
   public async loadAccountBalance(address: string, assetDenom: string): Promise<string> {
-    const response = await this.tendermintClient.getBalance(address, assetDenom);
-    return response?.amount ?? '0';
+    const response = await this.cosmosClient.get<BalanceResponse>(
+      `/cosmos/bank/v1beta1/balances/${address}/${assetDenom}`,
+    );
+    const balanceData = response?.data;
```
cdc-Hitesh, line 71: May be we can introduce `denom` equality check with our `assetDenom`, just to be safe.

src/service/rpc/NodeRpcService.ts
```diff
@@ -64,8 +65,11 @@ export class NodeRpcService implements INodeRpcService {
   }
 
   public async loadAccountBalance(address: string, assetDenom: string): Promise<string> {
-    const response = await this.tendermintClient.getBalance(address, assetDenom);
-    return response?.amount ?? '0';
+    const response = await this.cosmosClient.get<BalanceResponse>(
+      `/cosmos/bank/v1beta1/balances/${address}/${assetDenom}`,
```
cdc-Hitesh, line 69: Sorry for the basic question,
Does this       `/cosmos/bank/v1beta1/balances/${address}/${assetDenom}`, can fetch from any Cosmos compatible blockchain?

cdc-Hitesh (commented):
Left a few doubts.

src/service/rpc/NodeRpcService.ts
crypto-3dd: You see, the data returns by the endpoint does already the filtering. In fact, the caller even specifies the `denom`, so technically there's no need for another `denom` check, it returns a single object and that would just be redundant.

https://testnet-croeseid-4.crypto.org:1317/cosmos/bank/v1beta1/balances/tcro1lnj8dc9qazmlhm5t8vw8vryxyzv09zgep2cz9v/basetcro

src/service/rpc/NodeRpcService.ts
crypto-3dd: It's a standard endpoint for all cosmos-based chains like CRO, what's the difference is the base URL. Cro base URL will be different than let's say ATOM/COSMOS or any other forks.

You can check all other methods in that class to see where the same pattern is being re-used.

src/service/rpc/NodeRpcService.ts
cdc-Hitesh: okay sounds cool! 

Marking it resolved

src/service/rpc/NodeRpcService.ts
cdc-Hitesh: SUper super.
That is what I wanted to confirm.

Resolving it.

cdc-Hitesh (approved):
I assume the base branch is intentionally `dev` .
Rest all good @crypto-3dd

crypto-3dd:
> I assume the base branch is intentionally `dev` .
> Rest all good @crypto-3dd

Yup that's because this fix is not related to EVM (Cronos) integration

PR #659, opened 2021-08-31
Problem: Support multiple assets and non USD fiats

### LINKED ISSUE #660: Problem: Support multiple assets and fiat currency rate
Support coinbase v2 api to fetch market rates

cdc-Hitesh:
Fixes #660


src/service/rpc/MarketApi.ts
```diff
@@ -1,36 +1,79 @@
...
+      if (assetSymbol.toUpperCase() !== loadedSymbol || loadedCurrency !== currency.toUpperCase()) {
+        throw TypeError('Could not find requested market price info');
+      }
+
+      return {
+        assetSymbol: loadedSymbol,
+        currency: loadedCurrency,
+        dailyChange: croMarketPrice.data.coin.percent_change_native_24h,
+        price: croMarketPrice.data.coin.price_native.amount,
+      };
```
crypto-3dd, line 45: Hey @cdc-Hitesh  Not sure I'm following this part well, is this because `cro` price is not on Coinbase API?

Can we use a single API for everything directly?

cdc-Hitesh: I had 2 reasons to keep it the above way:
- to keep backward compatibility 
- Yes, `coinbase` can be used for `cro` as well but I wanted to keep CRO source from our own APIs

We can take a final call on this one, I personally would prefer single source.

crypto-3dd: I think it's better to keep it simple, I would just use a single source. 
When reading this code I would think there's something extra that's being handled, which is not happening. So it's extra complexity that might be misleading to other code readers. 🤔

cdc-Hitesh: I agree 100% on this. 
Switched our rates to be fetched from `coinbase` now.

crypto-3dd:
@cdc-Hitesh the test case is failing.

You can check it locally by running `yarn test`

crypto-3dd (approved):

PR #667, opened 2021-09-03
[v0.5.0] Internal Release

crypto-3dd:
### Additions
- Multi-Assets support added to the wallet
- EVM building blocks and CRONOS testnet support
- Multiple fiat currency conversion integration
- Support for Croeseid Testnet v4  
- Validator selection : UX update to reduce concentration of liquidity to  a few validators


cdc-Hitesh (commented):
- Package version change required?

src/utils/TransactionUtils.ts
```diff
@@ -89,21 +89,4 @@ export class TransactionUtils {
       },
     });
   }
-
-  // public static maxMinValidation(currentBalance: string, inputValue: string) {
-  //   const currentBalanceAmount = Big(currentBalance);
```
cdc-Hitesh, line 94: comment block

crypto-3dd: That was actually removed lol 😆

crypto-matto (approved):

PR #679, opened 2021-09-07
Fix QA issues reported

crypto-matto:
- [x] Fix display of config settings not updated after `Save`
- [x] Fix incorrect explorer url for Address explorer page
- [x] Small update from Design feedback


src/pages/assets/assets.tsx
```diff
@@ -272,8 +270,9 @@ const AssetsPage = () => {
           data-original={text}
           target="_blank"
           rel="noreferrer"
-          href={`${session.activeAsset?.config?.explorerUrl ||
-            session.wallet.config.explorerUrl}/tx/${text}`}
+          href={`${explorerUrl}/${
+            explorerUrl.indexOf('cronos') !== -1 ? 'address' : 'account'
```
crypto-3dd, line 274: Can you move the full URL to the configuration?

This is again not very flexible, because we are telling the front-end to know what **cronos** is. 

This reduces flexibility in case we need to add new assets, are we going to add other if/else here ? That wouldn't be very scalable.

It's preferable to update the config value so that it has **/address or /account in its url**. Front-end shouldn't do these things

crypto-3dd (changes requested):

src/pages/assets/assets.tsx
crypto-matto: Well said. I'll revise the Explorer URL mechanism.

crypto-3dd: I suggested you add all the necessary bits inside the config explorer.

For instance instead of     explorerUrl: `'https://cronos-explorer.crypto.org',` we would just add the` /address` or `/account `bits.

That way the front-end will remain fairly simple.

crypto-matto: I'm suggesting `explorerUrl` to change into an object:
```
explorerUrl: {
    home: 'https://crypto.org/explorer/',
    tx: 'https://crypto.org/explorer/tx',
    address: 'https://crypto.org/explorer/account',
    validator: 'https://crypto.org/explorer/validator',
}
```
Yet how to update all users’ existing config settings which were storing explorerUrl as string?

crypto-3dd: I see, wonder if we need the `home` explorer section?

> Yet how to update all users’ existing config settings which were storing explorerUrl as string?

They will be recreating the assets via the popup we provide, which would bring newly updated config object

src/config/StaticAssets.ts
```diff
@@ -18,9 +18,29 @@ const checkIfTestnet = (network: Network) => {
...
+      explorerUrl = DefaultWalletConfigs.TestNetCroeseid3Config.explorerUrl;
+      break;
+
+    case 'testnet-croeseid-4':
+      explorerUrl = DefaultWalletConfigs.TestNetCroeseid4Config.explorerUrl;
+      break;
+
+    default:
+      explorerUrl = DefaultWalletConfigs.MainNetConfig.explorerUrl;
+  }
```
crypto-3dd, line 39: This switch might not be needed because `walletConfig` argument comes already with the `walletConfig.explorerUrl`

We can just pass it directly.

src/pages/settings/settings.tsx
```diff
@@ -759,7 +759,7 @@ const FormSettings = () => {
       ...session.activeAsset!,
       config: {
         ...previousAssetConfig!,
-        explorerUrl: settingsDataUpdate.indexingUrl!,
+        // explorerUrl: settingsDataUpdate.indexingUrl!,
```
crypto-3dd, line 762: Is there a reason why we comment out this field?

crypto-matto: Indexing Url tends to be having `api` at the back, e.g. `https://crypto.org/explorer/croeseid4/api/v1/`, `https://cronos-explorer.crypto.org/api`, where `api/v1` and `api` exists.

src/pages/assets/assets.tsx
crypto-matto: I changed from `home` to `index` just to be clearer. The index page will be the same as the original `explorerUrl`

src/pages/settings/settings.tsx
crypto-matto: Now I update `explorer` instead

src/config/StaticAssets.ts
```diff
@@ -14,13 +14,50 @@ const checkIfTestnet = (network: Network) => {
...
+        break;
+      case 'validator':
+        url = `${explorerUrl}/validator`;
+        break;
+      default:
+        url = `${explorerUrl}`;
+    }
+  }
+  return url;
+};
```
crypto-3dd, line 51: I think rendering-related functions shouldn't be in static assets file, this file doesn't know about rendering, it only holds assets configuration.

src/config/StaticAssets.ts
```diff
@@ -14,13 +14,50 @@ const checkIfTestnet = (network: Network) => {
...
+      default:
+        url = explorer.index;
+    }
+  } else {
+    switch (page) {
+      case 'tx':
+        url = `${explorerUrl}/tx`;
+        break;
+      case 'address':
+        url = `${explorerUrl}/${explorerUrl.indexOf('cronos') !== -1 ? 'address' : 'account'}`;
```
crypto-3dd, line 41: I think this hardcoding here of `cronos` might still be very problematic, what if we have new assets added ? are we expanding into another large if/else switch statement?

I would suggest just declare simple configurations and each asset will know how to handle their own configuration

crypto-3dd (changes requested):

src/layouts/home/home.tsx
```diff
@@ -233,6 +234,46 @@ function HomeLayout(props: HomeLayoutProps) {
...
+      return;
+    }
+
+    setTimeout(async () => {
+      if (!walletSession.wallet.config.explorer) {
+        const updateExplorerUrlNotificationKey = 'updateExplorerUrlNotificationKey';
+
+        const allWallets = await walletService.retrieveAllWallets();
+
+        allWallets.forEach(async wallet => {
```
crypto-3dd, line 248: I like this approach 👍  we would need to test it greatly to make sure it's doing what we expect it to do exactly

src/pages/assets/components/ReceiveDetail.tsx
```diff
@@ -70,9 +70,12 @@ const ReceiveDetail: React.FC<ReceiveDetailProps> = props => {
 
   const assetAddress = (asset, _session) => {
     const { assetType, address } = asset;
-    // For IBC assets
     const { wallet } = _session;
-
+    // For Ledger
+    if (wallet.walletType === LEDGER_WALLET_TYPE) {
+      return wallet.address;
+    }
```
crypto-3dd, line 77: This would mean every time Ledger address would be only the `croxxx` type.

I think you can put a TODO over there, this section would be updated once @leejw51crypto brings support for Ledger on CRONOS

crypto-3dd (changes requested):

crypto-3dd (approved):

PR #681, opened 2021-09-07
Problem (Fix #251): design enhancement for undelegation

### LINKED ISSUE #251: Problem: design enhancement for undelegate
image

how about round-shaped box for `undelegate` ?

crypto-matto:
- [x] Create NodeRpcService fetching unbonding delegations 
- [x] Store the retrieved unbonding delegations info
- [x] Popup & layout for listing unbonding delgations
螢幕截圖 2021-09-13 下午1 17 06


src/pages/home/home.tsx
```diff
@@ -280,6 +280,7 @@ const HomePage = () => {
 
   useEffect(() => {
     const syncAssetData = async () => {
+      await walletService.syncAll();
```
crypto-3dd, line 283: Calling this `await walletService.syncAll();` every time the home page is opened would slow down lots of things

Can we maybe syncAll without forcing `awaiting` if that's extremely needed... I'm just worried that would create too much delay time and lots of requests every time the home page is opened which is not really the desired behavior.

crypto-3dd (changes requested):

crypto-3dd (approved):

PR #682, opened 2021-09-07
#672/ Chain indexing support for total claimed rewards

### LINKED ISSUE #672: Problem: Support user claimed total rewards in CRO
- Get total rewards claimed by an account on CRO Mainnet
- https://crypto.org/explorer/api/v1/accounts/cro1gaf3jqqzvrxvgc4u4vr6x0tlf6kcm703zqa34a/messages?pagination=offset&page=1&memo=&limit=1&order=height.desc&filter.msgType=MsgWithdrawDelegatorReward

cdc-Hitesh:
Fixes #672

### DIFF
--- a/src/service/rpc/ChainIndexingAPI.ts
+++ b/src/service/rpc/ChainIndexingAPI.ts
@@ -1,4 +1,5 @@
 import axios, { AxiosInstance } from 'axios';
+import Big from 'big.js';
 import {
   NftAccountTransactionListResponse,
   NftDenomResponse,
@@ -20,7 +21,6 @@ import {
 import { DefaultWalletConfigs } from '../../config/StaticConfig';
 import { croNftApi, MintByCDCRequest } from './NftApi';
 import { splitToChunks } from '../../utils/utils';
-import Big from 'big.js';
 
 export interface IChainIndexingAPI {
   fetchAllTransferTransactions(
--- a/src/service/rpc/ChainIndexingAPI.ts
+++ b/src/service/rpc/ChainIndexingAPI.ts
@@ -261,6 +261,7 @@ export class ChainIndexingAPI implements IChainIndexingAPI {
 
     while (currentPage <= totalPages) {
 
+      // eslint-disable-next-line no-await-in-loop
       const delegatorRewardMessageList = await this.axiosClient.get<AccountMessagesListResponse>(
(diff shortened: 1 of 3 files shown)


crypto-3dd:
```
src/service/rpc/ChainIndexingAPI.ts
  Line 225:11:  'rewardMsgList' is never reassigned. Use 'const' instead  prefer-const
```

This might be failing the CI.

src/service/rpc/ChainIndexingAPI.ts
```diff
@@ -213,4 +215,55 @@ export class ChainIndexingAPI implements IChainIndexingAPI {
...
+    }
+  }
+
+  private async getDelegatorRewardMessageList(address: string) {
+
+    // Note: handle `pagination`
+    const delegatorRewardMessageList = await this.axiosClient.get<AccountMessagesListResponse>(
+      `accounts/${address}/messages?order=height.desc&filter.msgType=MsgWithdrawDelegatorReward`,
+    );
+
```
crypto-3dd, line 260: I think we would just go ahead and paginate it directly.

cdc-Hitesh: Do you mean, 
We should handle it now?
or 
This will be handled automatically?

@crypto-3dd

crypto-3dd: > We should handle it now?

Yes, I mean handling it now.

cdc-Hitesh: I have handled pagination now.

src/service/rpc/ChainIndexingAPI.ts
```diff
@@ -20,6 +22,7 @@ import { DefaultWalletConfigs } from '../../config/StaticConfig';
 import { croNftApi, MintByCDCRequest } from './NftApi';
 import { splitToChunks } from '../../utils/utils';
 import { UserAsset } from '../../models/UserAsset';
+import { accountMsgList } from './ChainIndexingModels';
```
crypto-3dd, line 25: `'./ChainIndexingModels';` is being imported multiple times. This makes the CI fail

crypto-3dd (dismissed):

src/service/rpc/ChainIndexingAPI.ts
cdc-Hitesh: Fixed and thanks for the tip Eddy.

My eslint seems to not fix this 😢 , not sure what's wrong

crypto-3dd (approved):

PR #702, opened 2021-09-21
IBC general support - WIP

crypto-3dd:


crypto-3dd:
The priority has been lowered for this ticket. We might come back on this later.

PR #703, opened 2021-09-21
#699: Add one-click lock support on Chain desktop wallet

### LINKED ISSUE #699: Problem: Support one-click lock on Chain desktop wallet

cdc-Hitesh:
Fixes #699 

Screenshots:
image

image

image


crypto-matto:
This feature works and sounds! There are several places I'd suggest you to update though:
- [ ] Delete the close button in the modal, since it's no use at all
- [ ] See if you can also turns the lock icon into azure color

Since I suspect `SessionLockModal` is really similar to `PasswordFormModal`, I suggest you could reuse most from `PasswordFormModal` instead of creating a new component with the codes inside nearly the same. That would cause less maintenance work in the future. 
There are 2 ways to do so: 
1. Keep using `PasswordFormModal`, but passing your newly added `props` like `mask`, `maskClose`, etc to the component 
2. You may still create a `SessionLockModal` component, but as a Higher Order Component (HOC), which is pretty much like the concept of `extends`

crypto-matto (changes requested):
This feature works and sounds! There are several places I'd suggest you to update though:
- [x] Delete the close button in the modal, since it's no use at all
- [ ] See if you can also turns the lock icon into azure color

Since I suspect `SessionLockModal` is really similar to `PasswordFormModal`, I suggest you could reuse most from `PasswordFormModal` instead of creating a new component with the codes inside nearly the same. That would cause less maintenance work in the future. 
There are 2 ways to do so: 
1. Simply keep using `PasswordFormModal`, but passing your newly added `props` like `mask`, `maskClose`, etc to the component 
2. You may still create a `SessionLockModal` component, but as a Higher Order Component (HOC), which is pretty much like the concept of `extends`

I think the HOC concept is a bit advanced for you at the moment, so I have pushed the code for your reference this time. You can see the redundant code in `SessionLockModal.tsx` is greatly reduced and `SessionLockModal.less` is not even necessary at all now. Though you may keep the less file for other components in the future if you are applying a very different styling.

On the other hand, you may also:
- [ ] Refine some codes in `PasswordFormModal`

crypto-3dd (commented):
Looks overall super clean !

One small suggestion I would add is to make the lock button same width with the component below just for more symmetry when looking at it.

cdc-Hitesh:
Thanks for the code push @crypto-matto . Really appreciate it a lot, that's something new I learned here.Woww.

I have noted the points you mentioned. Will try to do it.


@crypto-3dd  Noted your review as well. Shall fix and push.

crypto-matto:
It's now stuck at the unlocked successfully screen after I entered the password. 
Please fix the following:
- `setTimeout()` to auto close the lock screen popup. 
-  The lock button width is not aligned at the center horizontally, like the rest of the buttons on nav bar. 螢幕截圖 2021-09-23 下午3 20 35

crypto-matto (approved):

crypto-3dd (approved):

PR #716, opened 2021-09-29
Backend Implementation for bridge related transactions  - Part I

crypto-3dd:
- [ ] Crypto .org to Cronos
- [ ] Cronos to Crypto .org


crypto-matto:
```
export enum BridgeTransferDirection {
  CRONOS_TO_CRYPTO_ORG = 'CRONOS_TO_CRYPTO_ORG',
  CRYPTO_ORG_TO_CRONOS = 'CRYPTO_ORG_TO_CRONOS',
  ETH_TO_CRONOS = 'ETH_TO_CRONOS',
  CRONOS_TO_ETH = 'CRONOS_TO_ETH',
}
```
In the future when more bridges are added, isn't it gonna be too many combinations (2^n) to define & hard to maintain? 
Or is it because every bridge transferral could be very different, so that's why we'll have to separate each type of bridge transfer so specifically?

crypto-3dd:
> In the future when more bridges are added, isn't it gonna be too many combinations (2^n) to define & hard to maintain? Or is it because every bridge transferral could be very different, so that's why we'll have to separate each type of bridge transfer so specifically?

A few clarifications here : 

- Every interaction between 2 chains will create 2 bridge transfer directions. There's no way around this, we still need to express this reality otherwise we are not bridging cross-chains.
- Those 4 directions are all we will ever need for the goal of implementing bridges on the desktop wallet app for the foreseeable future. Crypto .org to Cronos **(2 directions)** and Cronos to Crypto .org **(2 directions)**

PR #786, opened 2021-11-03
Problem (Fix #784): Potential endpoint format changes 

### LINKED ISSUE #784: Problem: Potential endpoint format changes
## Problem
The current CRONOS testnet public node endpoint will be keeping the same domain https://cronos-testnet-3.crypto.org with different ports
evm-rpc: https://cronos-testnet-3.crypto.org:8545
tendermint-rpc: https://cronos-testnet-3.crypto.org:26657
rest: https://cronos-testnet-3.crypto.org:1317
grpc: https://cronos-testnet-3.crypto.org:9090

Sadly though, we have some issues on worker scripts to the same domain.
For mainnet, we would like to split different domains instead of different ports, eg.
evm-rpc: https://evm-cronos.crypto.org
tendermint-rpc: https://rpc-cronos.crypto.org
rest: https://rest-cronos.crypto.org
grpc: https://grpc-cronos.crypto.org

If we're taking this approach, we shall facilitate this change in various places on Desktop Wallet, specifically on CRONOS mainnet.

crypto-matto:
May discuss whether we should support port auto filling or not for Testnet Node URL settings

### DIFF
--- a/src/service/rpc/clients/EVMClient.ts
+++ b/src/service/rpc/clients/EVMClient.ts
@@ -10,11 +10,11 @@ class EVMClient implements IEvmRpc {
   }
 
   public static create(web3HttpProviderUrlUser: string): EVMClient {
-    const words = web3HttpProviderUrlUser.split(':', 2);
-    const web3HttpProviderUrl = `${words[0]}:${words[1]}`;
-
-    if (!web3HttpProviderUrl.startsWith('https://') || !web3HttpProviderUrl.startsWith('http://')) {
-      const web3 = new Web3(new Web3.providers.HttpProvider(web3HttpProviderUrl));
+    if (
+      !web3HttpProviderUrlUser.startsWith('https://') ||
+      !web3HttpProviderUrlUser.startsWith('http://')
+    ) {
+      const web3 = new Web3(new Web3.providers.HttpProvider(web3HttpProviderUrlUser));
       return new EVMClient(web3);
     }
     throw new Error('Please provide a valid HTTP Web3 Provider.');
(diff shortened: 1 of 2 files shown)


src/service/rpc/clients/EVMClient.ts
```diff
@@ -12,7 +11,7 @@ class EVMClient implements IEvmRpc {
 
   public static create(web3HttpProviderUrlUser: string): EVMClient {
     const words = web3HttpProviderUrlUser.split(':', 2);
-    const web3HttpProviderUrl = `${words[0]}:${words[1]}${NodePorts.EVM}`;
+    const web3HttpProviderUrl = `${words[0]}:${words[1]}`;
```
crypto-eddy, line 14: We don't even need to process this URL anymore since we are just now using the whole `nodeUrl` string.

So we can remove the split section and pass directly `web3HttpProviderUrlUser` to the `Web3.providers.HttpProvider(web3HttpProviderUrlUser))` since no processing is needed further.

src/service/rpc/clients/EVMClient.ts
```diff
@@ -12,7 +11,7 @@ class EVMClient implements IEvmRpc {
...
 
     if (!web3HttpProviderUrl.startsWith('https://') || !web3HttpProviderUrl.startsWith('http://')) {
       const web3 = new Web3(new Web3.providers.HttpProvider(web3HttpProviderUrl));
```
crypto-eddy, line 17:
```suggestion
      const web3 = new Web3(new Web3.providers.HttpProvider(web3HttpProviderUrlUser));
```

crypto-eddy (approved):

PR #796, opened 2021-11-08
Fix incorrect explorer address url

crypto-matto:


XinyuCRO (approved):

crypto-eddy:
You can also update to correct testnet bridge indexing url `DefaultTestnetBridgeIndexingUrl` in the `BridgeConfig.ts` file.

PR #808, opened 2021-11-11
Feature (Fix #805): Allow users to specify manually destination address during bridge transactions

### LINKED ISSUE #805: Bridge : Allow users to specify manually destination address during bridge transactions
- Update the UI to present a destination address UI input. (@crypto-matto )

- Handle the custom UI address input on the backend bridge service. ( @crypto-eddy )

We can use this https://cronos.crypto.org/bridge as reference.

crypto-matto:
- [x] Frontend able to input a custom destination address
- [x] Validation & UX on custom destination address
- [x] Backend integration for custom destination address

May have a better suggestion on the UI
Screenshot 2021-11-11 at 12 52 46 PM


src/pages/bridge/components/CronosBridgeForm.tsx
```diff
@@ -492,6 +527,73 @@ const CronosBridgeForm = props => {
...
+                      setToAddress(cronosAsset?.address);
+                      break;
+                    }
+                    default:
+                  }
+                }
+              }}
+              className="disclaimer"
+            >
+              I want to send funds to another destination address.
```
crypto-eddy, line 566:
```suggestion
              I want to send funds to a custom destination address.
```

src/pages/bridge/components/CronosBridgeForm.tsx
```diff
@@ -138,19 +145,29 @@ const CronosBridgeForm = props => {
 
     switch (bridgeTo) {
       case 'CRYPTO_ORG': {
+        destinationAddress = tendermintAddress;
         setToAddress(tendermintAddress);
```
crypto-eddy, line 149: Looks like `destinationAddress` and `setToAddress` are being used for the same purpose.
Why not just use the address being set in `setToAddress`. It looks to me as a duplicate logic.

crypto-eddy (changes requested):

src/pages/bridge/components/CronosBridgeForm.tsx
crypto-matto: The `setToAddress` function is not updating `toAddress` immediately, and it doesn't have a callback, so we need an extra variable to update the formField value. 

Similar approach is often used in `Session` as well, where we first retrieve the latest session, store it in a variable, then update the session state
```
    const currentSession = await walletService.retrieveCurrentSession();
    const currentAsset = await walletService.retrieveDefaultWalletAsset(currentSession);

    setSession(currentSession);
```

Instead of update the session state right a way, then pass it to function.

```
    const currentSession = await walletService.retrieveCurrentSession();
    setSession(currentSession);
    const currentAsset = await walletService.retrieveDefaultWalletAsset(session);
```

src/components/RowAmountOption/RowAmountOption.tsx
```diff
@@ -7,16 +7,20 @@ import { scaledBalance, UserAsset } from '../../models/UserAsset';
 interface ValidatorPowerPercentBarProps {
   form: FormInstance;
   walletAsset: UserAsset;
+  setSendingAmount?: (value: string) => void;
```
XinyuCRO, line 10: event callback props can start with `on`, eg `onSetSendingAmount` for better convention, but no big deal here

XinyuCRO (approved):
LGTM

src/pages/bridge/components/CronosBridgeForm.tsx
crypto-eddy: The difference is we are not mutating the session.

But here we have a `let destinationAddress` mutable variable being mutated several places. This type of coding is usually very prone to mistakes and bugs.

Also I'm seeing the same being repeated a couple of places. Try to see if there's a way to remove duplications.

crypto-eddy (changes requested):

src/components/RowAmountOption/RowAmountOption.tsx
crypto-matto: This is not a callback props. This is really used to update the `sendingAmount` state outside of its parent component. 
The event callback you're referring is `onAmountOption()` - an `onClick()` like function instead.
e.g. when `amount` option is clicked, do something. `setSendingAmount()` is also called due to this event callback.

XinyuCRO: what I mean callback is something in the child happened, notify outer, let the outer decide what to do,  in this context, it's just like `onClick`.  `setSendingAmount` sounds like an action, not an event name, no big deal here, just some naming convention

btw, child component shouldn't change parent component's state, it's an anti-pattern

XinyuCRO (approved):

crypto-eddy (approved):

PR #809, opened 2021-11-11
Problem: New Transaction History Model

cdc-Hitesh:
Hey @crypto-eddy @crypto-matto @XinyuCRO 

Please help me validate this approach.
What are your thoughts?


src/service/rpc/ChainIndexingAPI.ts
```diff
@@ -379,4 +379,42 @@ export class ChainIndexingAPI implements IChainIndexingAPI {
 
     return finalMsgList;
   }
+
+  /**
+   * 
+   * @param userAddress Supports only Crypto.org USER addresses
+   * @param optionalMsgTypeName {Optional} Cosmos MsgType Name
+   */
+  public async getMessagesByAccountAddress(userAddress: string, optionalMsgTypeName?: string) {
```
XinyuCRO, line 388: we can expose `page` & `pageSize` param to this function and let the UI layer (eg. table) choose which page to load, not to load all txs at one time. If the user has too many txs, it will take quite a long time to fetch, bad for performance and UX

src/service/rpc/ChainIndexingAPI.ts
```diff
@@ -379,4 +379,42 @@ export class ChainIndexingAPI implements IChainIndexingAPI {
...
+    const finalMsgList: accountMsgList[] = [];
+
+    if (optionalMsgTypeName && SUPPORTED_MSGTYPE_NAMES_CHAIN_INDEXING.includes(optionalMsgTypeName)) {
+      queryURL = `accounts/${userAddress}/messages?order=height.desc&filter.msgType=${optionalMsgTypeName}&page=${currentPage}`;
+    }
+
+    while (currentPage <= totalPages) {
+      // eslint-disable-next-line no-await-in-loop
+      const messageList = await this.axiosClient.get<AccountMessagesListResponse>(
+        queryURL
```
XinyuCRO, line 403: why request the same URL in this loop?

src/service/rpc/ChainIndexingAPI.ts
XinyuCRO: also we can make `optionalMsgTypeName` a type, the caller can get a hint when calling this function from the type list

```ts
type MesgTypeName = "MsgSend" | "Msgxxx"
```

src/models/Transaction.ts
```diff
@@ -64,7 +66,28 @@ export enum NftTransactionType {
...
+  assetType: UserAssetType;
+}
+
+export interface BaseCommonTransaction {
+  walletId: string;
+  assetId?: string;
+  assetType: UserAssetType;
+}
+
+export interface CommonTransaction extends BaseCommonTransaction {
```
XinyuCRO, line 84: for the model definition, we can use a typescript technique union https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#discriminating-unions 

```ts
export interface CommonAttributesByWallet {
  customParams?: { [key: string]: string }; // Note: We WILL dedicate a storage for `customParams` like totalBalance, totalRewards, estimatedApy etc.
  walletId: string;
  assetId?: string;
  assetType: UserAssetType;
}

export interface BaseCommonTransaction {
  walletId: string;
  assetId?: string;
  assetType: UserAssetType;
}

export interface StakingTransactionModel extends BaseCommonTransaction {
  walletId: string;
  assetId?: string;
  txType: "staking"
  data: StakingTransactionData;
}

export interface UnbondingDelegationModel extends BaseCommonTransaction {
  walletId: string;
  assetId?: string;
  txType: "unbonding"
  data: UnbondingDelegationData;
}

export type CommonTransaction =
  | StakingTransactionModel
  | UnbondingDelegationModel
```

for the persistence layer, it can only accept `CommonTransaction` type, don't care the inner structure,
for the caller, it will be safer & easier to know the inner type

```ts
  let tx: CommonTransaction = {...}

  if (tx.txType === 'staking') {
    // we can sure this is type safe
    tx.data.stakedAmount
  }

```

src/service/rpc/ChainIndexingAPI.ts
cdc-Hitesh: > also we can make optionalMsgTypeName a type, 

Yes, I agree with you. That's why I have pulled all the supported message types in `staticConfig`. Making it a type that make it fail safe.


>we can expose page & pageSize param to this function and let the UI layer (eg. table) choose which page to load, not to load all txs at one time.

This is a good thought on performance! Not sure if this is a good time to introduce this feature . May be separate ticket for this?  cc @crypto-matto  What's your say?

src/service/rpc/ChainIndexingAPI.ts
cdc-Hitesh: Yes, correct.
 It's a work in progress function. 
I should have marked it as `TODO`. Thanks for noticing

src/models/Transaction.ts
cdc-Hitesh: Hmm. 
That is an interesting approach.
Just wondering if keeping a single model with a dedicated type for `txType` behaves well.

Like `txType: "staking" | "transfer"`

src/service/rpc/ChainIndexingAPI.ts
crypto-matto: For our current setup, all the transaction data are all stored in local storage through the below process:
`syncAll()` => `syncTransactionsData()` => `fetchAndUpdateTransactions()` => `fetchAndSaveTransfers()` => `fetchAllTransferTransactions()` => `saveTransfers()`
Therefore, the issue @XinyuCRO was referring might not exactly be the case as we don't need to call the api every time we change pages on the UI. We're just getting these data from local storage.

However, syncing time could be quite lengthy when users first open their app, as we're loading all data in one batch & update our local storage. I do agree with @XinyuCRO suggestion on exposing `page` & `pageSize` in the ChainIndexingAPI for future use, just that we may still need a certain set of parameter input (may be `page: 1` & `pageSize: 0`? ) to return us all transactions for the moment. 

In a separate PR in the future, we may revise the way we insert data in local storage, rather than everytime we retrieve all transactions data & reinsert the whole thing, we may try to fetch only latest transactions. That I believe could drastically reduce our current app loading time.

cdc-Hitesh:
cc @XinyuCRO @crypto-matto @crypto-eddy 
Please review the logic design for the new model.
I will have to do some extra finishing for this, as this is becoming a large piece of code.
Would request your thoughts, will continue once I come back from the vacation.
Thanks

cdc-Hitesh:
cc @XinyuCRO @crypto-matto @crypto-eddy 
Please review the logic design for the new model.

src/storage/StorageService.ts
```diff
@@ -402,9 +435,18 @@ export class StorageService {
...
+    const rewardCustomParams = await this.db.commonAttributeStore.findOne<RewardAttributesRecord>({
+      walletId: walletId,
+      type: 'reward'
+    });
+
     return {
       transactions: rewardTxs.map(tx => tx.txData),
-      walletId
+      walletId,
+      totalBalance: rewardCustomParams.customParams.totalBalance,
```
crypto-matto, line 462: The app is crashing as `rewardCustomParams.customParams` is `null`. 
All customParams using `this.db.commonAttributeStore.findOne()` could potentially causing this issue. 

Should add `|| '0'` at the back of each assignment as well. 
e.g. 
``` totalBalance: rewardCustomParams.customParams.totalBalance || '0' ```

**Review by crypto-matto — CHANGES_REQUESTED** [review]

File: src/models/Transaction.ts (hunk 3 above)
Comment (crypto-matto) [reply src/models/Transaction.ts]: I'm seeing you using the TypeScript Unions 👍🏻.

File: src/storage/StorageService.ts [hunk 5]
```diff
@@ -338,11 +383,31 @@ export class StorageService {
   }
 
   public async retrieveAllStakingTransactions(walletId: string) {
-    return this.db.stakingStore.findOne<StakingTransactionList>({ walletId });
+    let stakingTxRecord = await this.db.commonTransactionStore.find<StakingTransactionRecord>({
```
crypto-matto: I find there are lots of `...tx` variables using `let` yet isn't gonna change in values. Use `const` instead.

src/models/Transaction.ts
```diff
@@ -68,46 +68,82 @@ export enum NftTransactionType {
...
+  type: 'nft',
+  customParams: {
+    nftQuery?: NftQueryParams
+  }
+}
+
+export interface stakingAttributes extends CommonAttributesByWalletBase {
+  type: 'staking',
+  customParams: {
+    totalBalance?: string
```
crypto-matto, line 92: May be we should include `unbondingBalance` as well?

src/config/GeneralConfig.ts
```diff
@@ -5,4 +5,5 @@ export interface GeneralConfig {
   languageCode: string;
   isAppLockedByUser: boolean;
   incorrectUnlockAttempts: number;
+  // isTxHistoryMigrated: boolean;
```
crypto-eddy, line 8: Can be cleaned up if not needed

src/storage/StorageService.ts
```diff
@@ -535,12 +701,42 @@ export class StorageService {
...
+  public async insertCommonTransactionRecords(records: CommonTransactionRecord[]) {
     if (records.length === 0) {
       return Promise.resolve();
     }
-    return this.db.commonTransactionStore.insert<CommonTransactionRecord[]>(records);
+
+    records.forEach(async record => {
+
+      console.log(`[insertCommonTransactionRecords] Upserting record, txHash : ${record.txHash}`)
+      await this.db.commonTransactionStore.update<CommonTransactionRecord>({
```
crypto-matto: Is it possible to update in one single `update` command with an array, instead of calling multiple times in a forEach loop? 
Reference: https://docs.mongodb.com/manual/reference/method/db.collection.update/#modify-a-field-using-the-values-of-the-other-fields-in-the-document

src/storage/StorageService.ts
cdc-Hitesh: Yes correct noted.

src/storage/StorageService.ts
cdc-Hitesh: I agree!
I wanted `eslint` to take care of this for me 😆 (it's easy to see all such errors in a log) . But thanks for the notice.

src/models/Transaction.ts
cdc-Hitesh: I haven't seen any model using `unbondingBalance`, as I have referenced to current models. 
If you suggest we can keep that as well?

@crypto-matto

src/config/GeneralConfig.ts
cdc-Hitesh: Yes noted.
I have kept it to indicate a possible approach for data migration from old style to this latest style.

I will surely get rid of such comments.

src/storage/StorageService.ts
cdc-Hitesh: You are absolutely correct on this.
But the design of data updating is such that we need to call one by one.

If it `inserted` transactions, we can do batch calls.
For update I will have to figure out, will check your article.

src/storage/StorageService.ts
cdc-Hitesh: Resolved.

src/storage/StorageService.ts
```diff
@@ -534,4 +744,58 @@ export class StorageService {
...
+  // eslint-disable-next-line
+  public async insertCommonTransactionRecords(records: CommonTransactionRecord[]) {
+    if (records.length === 0) {
+      return Promise.resolve();
+    }
+
+    const upsertRequests = records.map(async record => {
+      
+      // eslint-disable-next-line no-console
+      console.log(`[insertCommonTransactionRecords] Upserting record, txHash : ${record.txHash} | TxType: ${record.txType}`)
```
XinyuCRO, line 763: this line causes too much `console.log` when user has too many TXs

src/storage/StorageService.ts
```diff
@@ -337,53 +391,150 @@ export class StorageService {
...
+
+    const rewardCustomParams = await this.db.commonAttributeStore.findOne<RewardAttributesRecord>({
+      walletId,
+      type: 'reward',
+    });
+
+    return {
+      transactions: rewardTxs.map(tx => tx.txData),
+      walletId,
+      totalBalance: (rewardCustomParams && rewardCustomParams.customParams) ? rewardCustomParams.customParams.totalBalance : '0',
```
XinyuCRO, line 429: use lodash `_.get`

mannutech: Is using a third party library worth here? 
I mean in terms of size cost.

src/storage/StorageService.ts
mannutech: I can suppress that.

src/storage/StorageService.ts
XinyuCRO: oh, just use `xx?.xx` is fine, this line is too redundant

cdc-Hitesh: I agree. 
Let's change it to use `xx?.xx` format

cdc-Hitesh: Pushed

XinyuCRO (approved):

crypto-matto:
I think it's better to merge this after the CRC20 support is released, and we could earn more time in testing the support of this with CRC20 tokens as well.

crypto-matto:
Encountering this when all my staking balances are 0 & clicked `Delegation Management` in `Staking` Page
螢幕截圖 2021-12-24 上午10 22 30

cdc-Hitesh:
> ## Existing issues
> * [ ]  Encountering this when all my staking balances are 0 & clicked `Delegation Management` in `Staking` Page
> 
> 螢幕截圖 2021-12-24 上午10 22 30
> 
> * [x]  Transaction records are not sorted by DateTime in desc order
> 
> 螢幕截圖 2021-12-24 下午12 08 56
> 
> * [ ]  Wallet balances on Home page now takes extremely longer to load, when I'm switching to other pages and then come back
> 
> 螢幕截圖 2021-12-24 下午12 49 11
> 
> * [ ]  Transaction records are not consistent. Sometimes it is loading more, sometimes it is less. I guess there's so
> 
> 螢幕截圖 2021-12-24 下午1 04 08
> 
> 螢幕截圖 2021-12-24 下午1 01 32
> 
> * [ ]  If switching between different assets, and then an asset is first loading its transaction record, the transaction records will stay as the previous assets' records until it finishes loading. I believe we may add a load spinning icon or this
> 
> 螢幕截圖 2021-12-24 下午1 04 24

That's wonderful testing points @crypto-matto  .
Can I request the account private key to test them at my end?

crypto-matto:
## Breaking changes
- Ditch the loading of transaction histories for all assets in 1 single function call, since we now support much more assets. Replace `fetchAndSaveTransfers()` with `fetchAndSaveTransfersByAsset()`. The transaction histories are only loaded when users are inside a specific asset detail page. 
- Transaction histories will not be saved inside local storage

crypto-matto:
Hey @cdc-Hitesh, I have a few questions:

Since I'd like to include the MsgType in Assets Page -> Transaction Table field
1. Is it the function `getMessagesByAccountAddress()` could technically replace the original `fetchAllTransferTransactions()` API call? 
2. Am I able to input an array of string with multiple `optionalMsgTypeName` instead in `getMessagesByAccountAddress(userAddress: string, optionalMsgTypeName?: string)`
3. Could you also revise `Transaction.ts => type TransactionData` into:
```
export interface TransactionData {
  hash: string;
  memo: string | null;
  date: string;
  status: TransactionStatus;
  assetSymbol: string;
  msgTypeName?: string;
}
```
As such, inside `fetchAndSaveTransfersByAsset() => const transferTransactions` , we may use a more general `const transactions` instead, which includes multiple `msgTypeName`?

cdc-Hitesh:
> 1. Is it the function `getMessagesByAccountAddress()` could technically replace the original `fetchAllTransferTransactions()` API call?
> 2. Am I able to input an array of string with multiple `optionalMsgTypeName` instead in `getMessagesByAccountAddress(userAddress: string, optionalMsgTypeName?: string)`
> 3. Could you also revise `Transaction.ts => type TransactionData` into:

As replied on Slack,
1. Yes we can do that
2. Yes, the indexing api supports sending multiple msg types
3. Okay, we can include msgTypeName? in the interface.

crypto-matto:
Updated Layout on Assets => Transaction
Screenshot 2022-01-12 at 2 19 04 PM

crypto-matto (approved):

src/service/TransactionHistoryService.ts
```diff
@@ -402,7 +404,17 @@ export class TransactionHistoryService {
               status: evmTx.isError === '1' ? TransactionStatus.FAILED : TransactionStatus.SUCCESS,
             };
 
-            return transferTx;
+            const transferTxRecord: TransferTransactionRecord = {
+              walletId: currentSession.wallet.identifier,
+              assetId: currentAsset.identifier,
+              assetType: currentAsset.assetType,
+              txHash: evmTx.hash,
+              txType: "transfer",
```
cdc-Hitesh, line 412: I think we can extract out this labels `"transfer"` to a constant variable?
What do you say?

XinyuCRO (dismissed):

crypto-matto (dismissed):

crypto-matto (approved):

PR #820, opened 2021-11-16
Problem: Fix delegation list fetching | Handle 404 | Add pagination

### LINKED ISSUE #722: When you delegate usable amount, then earned fonts are "lost" on the transaction list
**Describe the bug**
I dont know if this is the correct place, but the transaction log for the stakereward is not present in the trans log, when you delegate an amount to a validator, when you have an "unclaimed" reward.
I can see that the Desktop wallet doesnt have all transactions (only MsgSend type), but I guess the issue would be the same if you had MsgDelegate and DelegateReward and would calculate the balance from that?..
If i'm off, just close the incident (since the full trans log is not present in the app, and therefore is not directly directed at this app)

**Environment and Version (please complete the following information):**
 - OS: Windows
 - Version 0.4.2

**To Reproduce**
The key is to:
Make sure there are:
1. unclaimed stakereward on a validator (lets say there is 14,12345678
2. ONLY stake from your usable balance, to the validator. (lets say you stake 8,00000000)
3. then your transaction log on the explorer and on mintscan will only show the transaction for the 8,0000000
4. Your 14,12345678 are added to your balance BUT not shown in the transaction list

**Expected behavior**

**Screenshots**

**Additional context**

### LINKED ISSUE #812: Problem: Potential incorrect delegation info after staking operations
## Problem 
The community team has passed over a weird issue - The user has made a few undelegations and they look successful in the explorer. But nothing changed on his desktop wallet UI. Clicking the refresh button doesn't solve the problem for the user. 
Community team reported similar issues in previous version before.

Address: `cro1c3p2v2r9lnsnh0z06dac5peq4rczy3jjj5x6fn`
Screenshots after undelegation:
image (3)
image (4)
image (2)

## Possible cause
The user is probably not performing undelegation on Desktop Wallet, as we cannot reproduce the issue if we perform all staking operations with Desktop Wallet only. They may be doing it from somewhere else, and the Desktop Wallet is not fetching the latest staking list if they didn't perform any staking-related operations from our side & we display list from local storage.

## Study
See whether could reproduce the issue by:
- Perform some staking in Desktop Wallet
- Perform undelegation from another wallet client (may be DeFi wallet)
- See if there are any validation info unmatch
- If so, try to revise on the validation fetching mechanism on our side

### LINKED ISSUE #815: Bug: Handle HTTP API returning  `404  response`
There are certain places in code, where handling 404 is necessary after fetching the HTTP API `cosmosClient`.

Eg: 
```
const response = await this.cosmosClient.get<UnbondingDelegationResult>(
      `/cosmos/staking/v1beta1/delegators/${address}/unbonding_delegations`,
    );

```

### LINKED ISSUE #819: Problem: Add pagination to `NodeRpcServices` methods

cdc-Hitesh:
Fixes #812 
Fixes #815 
Fixes #722 
Fixes #819


src/service/rpc/NodeRpcService.ts
```diff
@@ -128,45 +137,119 @@ export class NodeRpcService implements INodeRpcService {
...
+
+    let paginationKey: PaginationNextKey = null;
+    let totalSumFinal = 0;
+    const delegationTransactionListFinal: Array<StakingTransactionData> = [];
+
+    while (true) {
+
+      // Fetch paginated data
+      // eslint-disable-next-line no-await-in-loop
+      const [paginatedDelegationResponse, nextKey] = await this.fetchDelegationsByAddressPaginated(paginationKey, address);
```
XinyuCRO, line 149: if we don't pass the `paginationKey` here, will the API return the paginated result with `nextKey` or the whole delegation list?

cdc-Hitesh: It will return the API return the paginated result with `nextKey`

crypto-eddy (approved):

crypto-eddy (approved):

PR #840, opened 2021-11-24
feat: Dapp Browser

XinyuCRO:
Introduce the provider backend for integrating Dapp in the Desktop Wallet.

## Working Demo

The working demo can be found in the settings page's first tab, currently support connect to wallet, sendTXs

image

## Definitions

- Provider
    - A JavaScript object made available to a consumer, that provides access to Ethereum by means of a Client.
    - It's like a communication bridge between client and wallet. The client can access it with `window.ethereum` object, the wallet can receive messages from the provider.
- Client (Dapp)
    - An endpoint that receives Remote Procedure Call (RPC) requests from the provider and returns their results.
    - The client runs in a web container, the choice of the web container will be detailed later.
- Wallet (Desktop Wallet)
    - An end-user application that manages private keys, performs signing operations and acts as a middleware between the Provider and the Client.

This PR basically implement the provider and the web container which hosts clients

## Architecture

image

- Wallet and web containers run in different processes, web containers are webview tags, host client, and provider.
- Wallet and web containers talk through IPC message and webview's `executeJavascript` call
- The provider is injected into the web container, using webview's preload script
- Provider act as `window.ethereum` in client-side, Dapp can directly use this object for communications

### Web Container, `<iframe />` or `<webview />` ?

According to [EIP-1193](https://eips.ethereum.org/EIPS/eip-1193)  we have to inject a global `ethereum` object into the  Dapp's web container's `window` object to make the wallet communicate with the Dapp,

After a time of research,  I found some limitations of the `iframe` tag:

- injecting Javascript into an `iframe` will cause [Same-origin policy](https://en.wikipedia.org/wiki/Same-origin_policy).
    - One way to fix this problem is to disable `webSecurity` in [BrowserWindow](https://www.electronjs.org/docs/latest/api/browser-window), which is strongly against the electron's [safe rules](https://www.electronjs.org/docs/latest/tutorial/security#6-do-not-disable-websecurity)
- iframe and our BrowserWindow share the same process.
    - if any exception happens in the iframe will cause our main window to crash

`webview` doesn't have the problems above, it's easier to inject Javascript, and runs in a different process, just like a tab in the chrome browser, which is quite like the behavior in our app's usage.

Though the electron team doesn't recommend using `webview` because of the dramatic architectural changes, in this situation, we should definitely use `webview` other than `iframe`. 

Besides, `webview` provides [a rich amount of API](https://www.electronjs.org/docs/latest/api/webview-tag#methods), which is more usable than `iframe`

### `useIPCProvider`

a custom hook that setup communications between web container and wallet, also response for send TXs, sign messages, etc

### `DappBrowser`

the UI layer that contains `webview`, confirmation promotion, etc

### `preload.js`

the provider code, injected into Dapps, should be packaged separately

## Credits

- [https://eips.ethereum.org/EIPS/eip-1193](https://eips.ethereum.org/EIPS/eip-1193)
- https://github.com/trustwallet/trust-web3-provider


crypto-matto (dismissed):
LGTM

crypto-matto (approved):

PR #849, opened 2021-12-01
CRC20 backend - Signing Token Transfer transactions

crypto-eddy:


crypto-matto:
I'm seeing `mainnetSymbol` is always `CRO` for whatever assets. Is it intended?

crypto-eddy:
> I'm seeing `mainnetSymbol` is always `CRO` for whatever assets. Is it intended? Since we used `mainnetSymbol` in a lot of places on `price` & `value` field, so that we could also display a proper pricing for testnet TCRO.

The mainnet symbol should be the token symbol, I've fixed that issue, should be okay in the next commit.

src/service/TransactionHistoryService.ts
```diff
@@ -413,6 +458,85 @@ export class TransactionHistoryService {
...
+
+    await Promise.all(
+      assets.map(async asset => {
+        switch (asset.assetType) {
+          case UserAssetType.TENDERMINT:
+            break;
+          case UserAssetType.EVM:
+            await this.fetchCurrentWalletCRC20Tokens(asset);
+            break;
+          default:
```
crypto-matto, line 533: Missing case for CRC_20_TOKEN
```
case UserAssetType.CRC_20_TOKEN:
```
This would result in CRC_20_TOKEN assets falling into:
```
throw TypeError('Not supported yet');
```

src/service/TransactionHistoryService.ts
```diff
@@ -413,6 +458,85 @@ export class TransactionHistoryService {
...
+    console.log('CRC_20_PERSISTED_TOKENS', {
+      address,
+      size: newlyLoadedTokens.length,
+      newlyLoadedTokens,
+    });
+
+    return newlyLoadedTokens;
+  }
+
+  public async fechTokensAndPersistBalances(session: Session | null = null) {
```
crypto-matto, line 513: Typo `fetchTokensAndPersistBalances`

src/service/TransactionHistoryService.ts
```diff
@@ -428,6 +552,9 @@ export class TransactionHistoryService {
       return;
     }
 
+    // Fetch and update tokens balances
+    await this.fechTokensAndPersistBalances(session);
```
crypto-matto, line 556: When there's an asset fall into the `default case` in `fetchTokensAndPersistBalances()` and throwing an error
```
        switch (asset.assetType) {
          case UserAssetType.TENDERMINT:
            break;
          case UserAssetType.EVM:
            await this.fetchCurrentWalletCRC20Tokens(asset);
            break;
          default:
            throw TypeError('Not supported yet');
        }
```
This line will stuck in the `await` and anything after that will not run at all.

crypto-matto (changes requested):

src/service/TransactionHistoryService.ts
```diff
@@ -264,6 +264,51 @@ export class TransactionHistoryService {
...
+
+                const transferTx: TransferTransactionData = {
+                  amount: crc20TokenTx.value,
+                  assetSymbol: currentAsset.symbol,
+                  date: transactionTime,
+                  hash: crc20TokenTx.hash,
+                  memo: '',
+                  receiverAddress: crc20TokenTx.to,
+                  senderAddress: crc20TokenTx.from,
+                  status: TransactionStatus.SUCCESS,
```
cdc-Hitesh, line 295: Are we not checking the `status` value here?

src/service/TransactionSenderService.ts
```diff
@@ -147,6 +147,97 @@ export class TransactionSenderService {
...
+            currentAsset.config?.indexingUrl,
+          );
+
+          const transfer: TransferTransactionUnsigned = {
+            fromAddress,
+            toAddress: transferRequest.toAddress,
+            amount: String(scaledBaseAmount),
+            memo: transferRequest.memo,
+            accountNumber: 0,
+            accountSequence: 0,
```
cdc-Hitesh, line 171: shall we set `accountNumber` and `accountSequence` to something invalid as default like `-1`

src/service/TransactionSenderService.ts
```diff
@@ -147,6 +147,97 @@ export class TransactionSenderService {
...
+          const encodedABITokenTransfer = evmTransactionSigner.encodeTokenTransferABI(
+            currentAsset.contractAddress,
+            transfer,
+          );
+
+          const web3 = new Web3('');
+          const txConfig: TransactionConfig = {
+            from: currentAsset.address,
+            to: currentAsset.contractAddress,
+            value: web3.utils.toWei(transferRequest.amount, 'ether'),
```
cdc-Hitesh, line 184: `transferRequest.amount` , can it be `undefined` or `null`

src/service/TransactionSenderService.ts
```diff
@@ -147,6 +147,97 @@ export class TransactionSenderService {
...
+            data: encodedABITokenTransfer,
+          };
+
+          const prepareTxInfo = await this.transactionPrepareService.prepareEVMTransaction(
+            currentAsset,
+            txConfig,
+          );
+
+          const staticTokenTransferGasLimit = 130_000;
```
cdc-Hitesh, line 193: Can we pull this constant somewhere on the top?

src/service/signers/EvmTransactionSigner.ts
```diff
@@ -39,6 +43,59 @@ class EvmTransactionSigner implements ITransactionSigner {
...
+
+    const gasPriceBN = web3.utils.toBN(
+      transaction.gasPrice || transferAsset?.config?.fee?.networkFee!,
+    );
+    const encodedTokenTransfer = this.encodeTokenTransferABI(
+      transferAsset.contractAddress,
+      transaction,
+    );
+
+    const chainId = transaction?.asset?.config?.chainId || 338;
```
cdc-Hitesh, line 73: Can we pull `338` as a constant on top?

src/service/signers/EvmTransactionSigner.ts
```diff
@@ -39,6 +43,59 @@ class EvmTransactionSigner implements ITransactionSigner {
...
+    const txParams = {
+      nonce: web3.utils.toHex(transaction.nonce || 0),
+      gasPrice: web3.utils.toHex(gasPriceBN),
+      gasLimit: transaction.gasLimit || transferAsset?.config?.fee?.gasLimit,
+      to: transferAsset.contractAddress,
+      value: 0,
```
cdc-Hitesh, line 79: Some token transfers can have optional non-zero ethereum value as well. 
We may need to input `value` as well.

src/service/signers/EvmTransactionSigner.ts
```diff
@@ -39,6 +43,59 @@ class EvmTransactionSigner implements ITransactionSigner {
...
+    const signedTx = await ethers.Wallet.fromMnemonic(phrase).signTransaction(txParams);
+    return Promise.resolve(signedTx);
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  public encodeTokenTransferABI(
+    tokenContractAddress: string,
+    transaction: TransferTransactionUnsigned,
+  ) {
+    const web3 = new Web3('');
```
cdc-Hitesh, line 93: Is there a benefit to instantiating Web3 in every function?
We can expose a single instance.

cdc-Hitesh (commented):
Left a few comments

src/service/TransactionHistoryService.ts
crypto-eddy: This place is where we are loading tokens for each asset/coin type, CRC_20_TOKEN is already a token type, it's not needed here. Only main coins tokens are loaded over here.

src/service/TransactionHistoryService.ts
crypto-eddy: Good catch

src/service/TransactionSenderService.ts
crypto-eddy: For tokens transfer, recipient is usually null, and the actual transfer happens in the encoded tokens transfer data. I will update.

src/service/signers/EvmTransactionSigner.ts
crypto-eddy: This is just for the utility function access, not much use, it will get garbage collected when the function execution steps out of that method scope.

crypto-eddy:
I've now responded to all the review comments and I've pushed the necessary updates.

crypto-matto (approved):

crypto-matto:
I'm seeing that if one of my CRC20 token balance got cleared out (Let's say I all traded them all on VVS Swap), while my balances all the corresponding tokens are updated, my supposed-to-be cleared out token's balance is not updated & still there. May have to investigate why.

crypto-matto:
> I'm seeing that if one of my CRC20 token balance got cleared out, while my balances all the corresponding tokens are updated, my supposed-to-be cleared out token's balance is not updated & still there.
> 
> Let's say my balance is having 10 USDT, and I spent all buying VVS on VVS Swap
> 
> * VVS balance is updated correctly
> * CRO balance after paying gas fee is updated correctly
> * USDT balance stays at 10 USDT
> 
> May have to investigate why.

I’m safely say that even after 8 hours the supposed-to-be zero balance asset is not updated, while all my non-zero balances assets are all updated.

I guess it’s about the backend is not removing the zero balance asset from existing DB record.

crypto-matto:
### Issues to be solved
- [ ] Zero balances assets is not removing & stays when the old balance
- [ ] Settings -> Node Configurations hide all non-CRO assets settings

crypto-matto (approved):
LGTM

PR #882, opened 2021-12-15
#816 | Added Icon_url support

### LINKED ISSUE #816: Problem: Loading CRC token data by `address`
The RPC method should return contract data for CRC20 tokens provided the user address is supplied.

Questions:

- Do we need to fetch token balances on a Cronos address?
- Any other contract-specific data we need to fetch as well?

@crypto-eddy

cdc-Hitesh:
Fixes #816

### DIFF
--- a/src/config/StaticConfig.ts
+++ b/src/config/StaticConfig.ts
@@ -134,6 +134,24 @@ export interface SupportedCurrency {
   symbol: string;
   label: string;
 }
+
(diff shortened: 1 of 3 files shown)


XinyuCRO (dismissed):

crypto-matto:
I see that `WBTC` is missing from the official token list https://cronos.crypto.org/explorer/tokens. 
Other than that, everything looks good.

crypto-matto (approved):

PR #900, opened 2021-12-31
feat: CRC20 token whitelist support

XinyuCRO:
Currently only support CRC20 token list on Cronos main net, on other chains(eg. test net), there is no whitelist data, so we consider all tokens are whitelisted and no warnings have to be shown.

- UI
image
image
image


crypto-matto:
Overall is working great~ There are some suggestions:
1. I would suggest adding a warning message in Asset Detail as well. We may use the same message. 
2. Shall we consider merging both `CRC20_TOKEN_ICON_URL` & `CRC20_MAINNET_WHITELIST_TOKENS` into one `Map`? We may consult @cdc-Hitesh on where to add the Icon Url.

crypto-matto (dismissed):

src/config/CRC20Tokens.ts
```diff
@@ -0,0 +1,174 @@
+export interface CRC20TokenInfo {
+  address: string;
+  // is tradable on VVS
+  isWhitelist: boolean;
+  iconURL: string;
+}
+
+// https://cronos.crypto.org/explorer/tokens
+// don't know why all response from tokenlist returns lowercased contract address
+// it's definitely not correct to use lowercased contract address here
```
cdc-Hitesh, line 10: Can we use checksum addresses?

src/models/UserAsset.ts
```diff
@@ -69,6 +69,8 @@ export interface UserAsset {
   // CRC20/ERC20 assets should have a contract address persisted along side them.
   // This will be used later for all actions interacting with the token like tokens transfers, etc, ...
   contractAddress?: string;
+
+  isWhitelisted?: boolean;
```
cdc-Hitesh, line 73: Is it a good idea to maintain `isWhitelisted` flag  on the `UserAsset` level or maintain a list of whitelisted addresses?
Just Curious to know.
@XinyuCRO

src/config/CRC20Tokens.ts
```diff
@@ -0,0 +1,174 @@
+export interface CRC20TokenInfo {
+  address: string;
+  // is tradable on VVS
+  isWhitelist: boolean;
```
cdc-Hitesh, line 4:
```suggestion
  isWhitelisted: boolean;
```

cdc-Hitesh (commented):
PR is strong.
Left a few comments.

src/models/UserAsset.ts
XinyuCRO: Yeah, I believe it just likes `iconURL` and other properties, are Asset's attributes, and it's easy to use from outside

src/config/CRC20Tokens.ts
XinyuCRO: Good point!

cdc-Hitesh (approved):

crypto-matto (approved):
LGTM Good work

PR #903, opened 2022-01-04
Update App Icon

crypto-matto:
Revert the original App Icon from previous Christmas edition 
- #853


XinyuCRO (approved):
wen remove beta 🤔

crypto-matto:
> wen remove beta 🤔

Until Cronos Mainnet Chain also removes Beta. From previous consensus, by the time we launch Gravity Bridge, Beta would likely be removed.

PR #911, opened 2022-01-06
feat: Full-featured Dapp Browser

XinyuCRO:
- Support URL input and search Google with keywords
- Full-featured URL navigation, forward/backward/reload
- Bookmark favorite DApp
- Ledger Support
image
image
image


crypto-matto:
## Issues
1. `RequestConfirmation.tsx` => useEffect() => if (event.name === 'tokenApproval') => Confirmation URL is not updated with current url field
Untitled
2. Potential app crash when input amount = 0 / null / undefined from DApp requests
Untitled 2
Video: https://www.loom.com/share/8b23348f5de2496d810a41d4a298739b

crypto-matto (approved):
I'm seeing Bookmark(and Disclaimer record) stored in localStorage instead of our normally using NeDB. 

Would it be better (more persist) to store Bookmark data in NeDB?

PR #919, opened 2022-01-11
Problem: Fix DelayedVestingAccount Not supported issue

### LINKED ISSUE #879: Problem: Missing support of DelayedVestingAccount rewards withdrawal
## Problem
Missing support of [DelayedVestingAccount](https://docs.cosmos.network/master/core/proto-docs.html#cosmos.vesting.v1beta1.DelayedVestingAccount). This function would be used for CrossFire dryrun participants (around 200-300), which they are eligible for redeeming their rewards recently.
vestingAccount
- It is uncertain whether the function is deprecated or was not supported from the beginning

cdc-Hitesh:
Fixes #879 

#879 Updated `@cosmjs/stargate` to `0.27.0`


crypto-matto (approved):
Everything working as expected 👍🏻👍🏻

PR #992, opened 2022-02-18
feature: DApp Ranking

XinyuCRO:
## Background

Promote popular DApps to users and let them access DApps with Desktop Wallet more easily.

## Implementation

- Load hard-coded DApp list
- Fetch TVL and other data about DApps via Defi Llama API

## Screenshot

- DApp List
image

- Filter Categories
image

- Add a close button to the browser, so that users can close the browser without leaving the page, keeping a more continuous UX
image

- This PR also fixes a layout issue in governance detail page 
image


crypto-matto (commented):
Not much to say~ Thoughtful & nicely done! 

With a few suggestions:
1. Make `Name` the only clickable field to direct to the DApp & apply css `cursor: pointer`. If you enclose the whole `Name` field with `<a>` then it shall be already applying `cursor:pointer` & `color: #1199FA`
2. For `Audit`, we may use `<IconTick>` tick (reference `nft.tsx`) instead of `Y/N` for audited DApps. Keep the field blank for those non-audited.  



Also please fix the audit check failure.

crypto-matto (approved):

Issue #971, opened 2022-02-07
Experiment: Change the bundle system from webpack to Vite for easier maintainability and faster build 


XinyuCRO:
The current bundle system is webpack, several problems to addresses
- the bundle scripts are ejected from `create-react-app` which are hard to maintain or upgrade
- the production bundle time on M1 Pro chip is almost 1 minute, which is a little too long

[Vite](https://vitejs.dev/) is built on [esbuild](https://esbuild.github.io/) which is extremely fast on development & bundle time, it also supports many features out of the box, e.g. less, which means,
- bundle scripts are easier to maintain
- we can get faster dev and production build

The Vite itself is younger and may not **that** stable for now, will keep looking into it

PR #1011, opened 2022-02-28
feat: Vite integration for dev build

XinyuCRO:
Provide [Vite](https://vitejs.dev/) as an alternative to use during the development process.

Note: This PR **only** add another way of running the bundle & dev server to extend and speed up the current workflow during development. All the existing behavior(webpack dev) will exist, also this PR will **not** affect the release build.

### Background
https://github.com/crypto-com/chain-desktop-wallet/issues/971

### Benchmark
|                 | Vite | Webpack |
| --------------- | ---- | ------- |
| Cold Start Time | 16s  | 35s     |

* For Vite, make sure it's [Pre-Bundling](https://vitejs.dev/guide/dep-pre-bundling.html#dependency-pre-bundling) finishes(this step is done during the first dev run, after that it will use it as a cache to improve build efficiency)

Vite cuts the cold start time to about half compared to webpack.

## Other things to notice

- Change the way of using asset icons

From, 
```js
import icon from './icon.svg';

const Asset = {
  ...
  icon: icon
 }
```

To,
```js
const Asset = {
  icon: './assets/icon.svg'
}
```

This is mainly because we reference the asset icon in `StaticAsset` and save the build time generated URL into the Database, the URL will change based on bundle system, it's not a good practice to persist build time value. 
In stead, we treat the icons as static assets and put them in the `public` folder, the URL won't change.


crypto-matto:
The electron app pops out within 20 sec~ Amazing 👍🏻 It is like only 25-40% of the original compile time. Great work! 

Also thanks for the suggestions on Asset & Bridge icons. We shall definitely move towards your approach in future development regarding to any similar use case. 

Just 2 more things to add:
1. I'd prefer the command in `package.json` as `vite:dev` to be consistent with the rest
1. Update README.md mentioning the usage of the above command

crypto-matto (approved):

PR #1018, opened 2022-03-11
feat: add file nesting support for VSCode

XinyuCRO:
It's a newly supported feature on VSCode https://code.visualstudio.com/updates/v1_64#_explorer-file-nesting

We can get a cleaner file structure without actually moving files after setting it

image


crypto-matto (approved):
That's a good move. I'm seeing the nested folder structure after upgraded my VS Code to latest.

PR #1030, opened 2022-03-21
Problem (Fix #1012): No opt-in for auto-upgrader

### LINKED ISSUE #1012: Problem: No opt-in for auto-upgrader
## Background
We encounter a lot of occasions where the latest release is either having bug, or not working probably for some random specific users. While we'd recommend our users to downgrade the app at that moment, the auto-upgrader always popup & force the upgrade to proceed.

## Solution
- The auto upload shall be enabled by default
- The existing notification popup should include an `upgrade` button, and only start the upgrade process after the button is clicked
- Add an opt-in config settings to enable & disable the auto upgrader. 
- Add a setting which the upgrade is only temporarily suspended for 1 month, avoiding users never upgrade their app again

crypto-matto:
## Issue
- #1012

## Screenshot
s2
s1'


electron/main.ts
```diff
@@ -18,6 +16,8 @@ import { getGAnalyticsCode, getUACode, actionEvent, transactionEvent, pageView }
 (global as any).getUACode = getUACode;
 (global as any).getGAnalyticsCode = getGAnalyticsCode;
 
+const Store = require('electron-store');
```
XinyuCRO, line 19: use `import`

XinyuCRO (commented):
Is it necessary to store the config in two places`electron-store` & `nedb`?

crypto-matto:
> Is it necessary to store the config in two places`electron-store` & `nedb`?

Electron-store doesn’t look persist to me? From my trial, if I don’t pass the value in ‘layout/home/index.tsx’ in prior, seems the value in Store() will always be reinitiated.

XinyuCRO:
checked out to the branch, there is an exception which blocks the sync 
image

image

XinyuCRO:
> Electron-store doesn’t look persist to me? From my trial, if I don’t pass the value in ‘layout/home/index.tsx’ in prior, seems the value in Store() will always be reinitiated.

`electron-store` will save configs in files, in this case, `~Library/Application Support/chain-desktop-wallet/config.json` is used for persistence.

Also one key-value pair (`expireTime`) is enough for use, if it's empty or expire, check update right away, otherwise skip the check, easier to maintain

crypto-matto:
> > Electron-store doesn’t look persist to me? From my trial, if I don’t pass the value in ‘layout/home/index.tsx’ in prior, seems the value in Store() will always be reinitiated.
> 
> `electron-store` will save configs in files, in this case, `~/Library/Application Support/chain-desktop-wallet/config.json` is used for persistence.
> 
> Also one key-value pair (`expireTime`) is enough for use, if it's empty or expire, check update right away, otherwise skip the check, easier to maintain

Let me try it out 👌🏻

XinyuCRO (approved):

PR #912, opened 2022-01-06
Ethereum Integration

crypto-eddy:
## Ethereum Chain
- [x] Handle new static ETH asset creation.
- [x] Load new ETH asset balance
  - Indexing Endpoint: https://eth-indexing.crypto.org/ethereum/mainnet/api/v1/address/0x2992b824ed06a99d0816d39c77e9512424253efc/balance
- [x] Support transferring new ETH asset. 
  - Node RPC Endpoint: https://eth-indexing.crypto.org/ethereum/mainnet/rpc 
- [x] Support for Ledger.
- Indexing service on Transaction History is still not updated to latest block. Direct users to Etherscan instead. 

## Screenshots
Screen Shot 2022-01-06 at 11 00 32 AM
Screen Shot 2022-01-06 at 11 05 37 AM
螢幕截圖 2022-08-03 下午6 05 16

## Other Outstanding supported Features
- #1032 
- #1167 
- #1209

PR #1032, opened 2022-03-22
feat: Support EVM chains switch in DApp Browser

XinyuCRO:
Add add/switch EVM compatible support for DApp Browser.

## Things to notice
- Add support for `wallet_addEthereumChain` and `wallet_switchEthereumChain` Ethereum RPC methods, which mean we can handle  add/switch chain request  from DApps, eg. [Chainlist](https://chainlist.org/)
- Add a new section in DApp Browser to manage chain configs
- Add support for the fee market EIP-1559 implementation in every EVM chain transaction inside DApp Browser (we should support this for other operations in wallet later, since testnet has already got the fee market feature online)
- Change the user agent in DApp Browser back to the normal one(instead of faking to be a mobile device). 
	- the reason is because some DApps like UniSwap's connect wallet modal shows nothing when the environment is mobile, so we have to remove the useragent hack to make these websites to work. 
	- the side effect of removing this can cause some website's auto-connecting not work, but if user connect it by hand once, it will work as usual later.

## Screenshoots
- chain switch in addressBar
image

- chain config tab for users to manage chains
image

- Add/Switch chain request modal

<div>
image
image
</div>

- Confirm Tx drawer will show different currency symbol when using other chains
image


crypto-matto:
## Issues
- The `Remove` button on newly added chains is not clickable until some actions are done (unknown to me)
- May just hide `Remove` button for Cronos Mainnet & Cronos Testnet

XinyuCRO:
@crypto-matto 

>  The Remove button on newly added chains is not clickable until some actions are done (unknown criteria to me)
yeah, can not remove the chain if it is currently being used

>  Network, after approval, isn't switched on web frontend until wallet disconnect & reconnect manually
Which DApp particular? I tested with top DApps in cronos ranks page, they all works fine

crypto-matto:
> 



> @crypto-matto

> > Network, after approval, isn't switched on web frontend until wallet disconnect & reconnect manually
> 
> Which DApp particular? I tested with top DApps in cronos ranks page, they all works fine
> 
Pancake swap for instance. https://pancakeswap.finance/

crypto-matto:
> 



> @crypto-matto

> > Network, after approval, isn't switched on web frontend until wallet disconnect & reconnect manually
> 
> Which DApp particular? I tested with top DApps in cronos ranks page, they all works fine
> 
Pancake swap for instance. https://pancakeswap.finance/

XinyuCRO:
Will be merged with #912

PR #1052, opened 2022-04-13
feat: add password protection on Address Book

### LINKED ISSUE #1051: Problem: Missing Password protection on Address Book
## Background
It’s a security enhancement proposed by the Security Team, which is to avoid the tamper of the address book and causing unintended fund loss.

## Solution
- The address book itself no need to be encrypted. Just that it requires the correct password to edit.
- Requires password when create, edit or delete the addresses.

XinyuCRO:
Close #1051

- This PR introduce a convenient hook for using the PasswordModal, need only few lines code to show the modal and save password in memory for later usage.

```js
const { show: showPasswordModal } = usePasswordModal();

// if we already get the password, won't show it again, the onSuccess will callback at once 
showPasswordModal({
  onSuccess: () => {
    // do staff
  }
})
```

- Also update the DApps list on Cronos


src/components/PasswordForm/PasswordFormModal.tsx
```diff
@@ -87,81 +134,146 @@ const PasswordFormModal: React.FC<PasswordFormModalProps> = props => {
...
+
+    if (passphrase.length > 0) {
+      props.onSuccess(passphrase)
+      return;
+    }
+
+    if (isShowing) {
+      return
+    }
+    const mm = Modal.info({
```
crypto-matto, line 226: Shall avoid naming like this.

src/components/PasswordForm/PasswordFormModal.tsx
```diff
@@ -87,81 +134,146 @@ const PasswordFormModal: React.FC<PasswordFormModalProps> = props => {
...
     </Modal>
-  );
-};
+  )
+}
+
+
+const usePasswordModal = () => {
+
+  const [m, setM] = useState({
```
crypto-matto, line 205: Shall avoid naming like this. Please pick a readable name.

crypto-matto, line 205: Shall avoid naming like this. Please pick a readable name.

crypto-matto (commented):
Overall is working great~ Normal asset transfer is working properly & not affected as well.


### To be done
- Change to readable naming
- Fix audit

crypto-matto (approved):

PR #1053, opened 2022-04-13
Fix Audit

crypto-matto:


XinyuCRO:
I don't know whether it will break the app or not, it has highly chance that there are API break changes with `async` npm package from 0.2.4 to 3.2.2. 
Our dependency `@seald-io/nedb` has heavily use of this lib, need do a check on

crypto-matto:
So far no problems from my side. I checked some indexing services & made some transaction broadcasts. 
Or if QA further study & finds out any issues which seems unsolvable, we may roll back to ignore this audit issue.

crypto-matto:
So far no problems from my side. I checked some indexing services & made some transaction broadcasts. 
Or if QA further study & finds out any issues which seems unsolvable, we may roll back to ignore this audit issue.

XinyuCRO:
I would suggest to patch the changes here https://github.com/caolan/async/pull/1828 with  https://www.npmjs.com/package/patch-package in our codebase and ignore this vulnerability

`@seald-io/nedb` has so many use of `async` can't guarantee jump from 0.x to 3.x will not break anything.
But if the mapping doesn't break anything, I am ok with this change.

XinyuCRO (approved):

crypto-matto:
> I would suggest to patch the changes here [caolan/async#1828](https://github.com/caolan/async/pull/1828) with https://www.npmjs.com/package/patch-package in our codebase and ignore this vulnerability

This package looks new to me & sounds good to be a solution when we encounter packaging issues which we just couldn't ignore in the future 👍🏻

crypto-matto:
> I would suggest to patch the changes here [caolan/async#1828](https://github.com/caolan/async/pull/1828) with https://www.npmjs.com/package/patch-package in our codebase and ignore this vulnerability

This package looks new to me & sounds good to be a solution when we encounter packaging issues which we just couldn't ignore in the future 👍🏻

PR #1067, opened 2022-04-26
feat: gas customization

XinyuCRO:
Add custom gas options for Crypto.org & Cronos Chain

image
image
image
image


crypto-matto:
## Issues
- [ ] Tx failure on DApp Browser
er1
- [ ] `Save` button is not responding when settings are unchanged. May disable it when values are unchanged. 
er2

crypto-matto:
## Issues
- [ ] Tx failure on DApp Browser
er1
- [ ] `Save` button is not responding when settings are unchanged. May disable it when values are unchanged. 
er2

XinyuCRO:
>We can add a Default button next to Save or Cancel button. When clicks it, the default Standard values for Network Fee & Gas Limit will be filled in. Also Cancel button may be renamed to Discard Changes like in existing Config.

- Added a "reset changes" button to set gas settings and remove the "cancel" button,  users can dismiss the modal to discard, also the modal will be too crowded if we add three buttons
- DApp gas settings will not have the "reset" button cause all gas are calculated by the current network status
image


>The original Network Fee & Gas Limit config settings would be redundant, and not used in any other operations. Shall be removed? (Hide from frontend only or remove together with backend as well is worth to discuss)

They share the same settings config underneath, can leave it there

XinyuCRO:
>  the new values are not showing in Settings. They're still the old value.

Yeah, I've noticed that as well, and fixed it in the newer commits, can give it a try

crypto-matto (approved):
LGTM now

crypto-matto (approved):

Issue #959, opened 2022-01-27
Problem: Missing support of batch withdraw rewards


crypto-matto:
## Background
It could be troublesome to withdraw rewards from multiple delegators, especially if a user have plenty delegations. 

## Propose Solution
Add a `Withdraw All Rewards` button at the top right corner. The single transaction will be including multiple rewards withdrawal signed messages.

## Task
- [ ] A `WalletService` function to broadcast single transaction including multiple rewards withdrawal signed messages
- [ ] A CTA button `Withdraw All Rewards`
- [ ] A confirmation modal involving cumulative rewards info

PR #1088, opened 2022-05-16
Feature/#959/batch withdrawal frontend

yi-cro:
Fixes https://github.com/crypto-com/chain-desktop-wallet/issues/959

- [x] A CTA button Withdraw All Rewards
- [x] A confirmation modal involving cumulative rewards info


crypto-matto:
## Issue
- [ ] The `Withdraw All Rewards` modal is not displaying data correctly on first attempt, until open it twice. 
- [ ] Doubled `(s)(s)` in `Withdraw Reward(s) From Validators(s)(s)`. 
- [ ] The withdrawal validator list ordering, together with `Rewards` amount is changed after the `Confirm` button is clicked. 
## Suggestion
- [ ] `Withdraw All Rewards` button could remove the border, or use `type="primary"` instead, since the current button style is intended not to suggests users clicking on it (e.g. `Clear Storage` button in `Settings`) 
- [ ] You could definitely make good use of the existing `REWARD BALANCE` at the top to display total rewards

crypto-matto:
Like this design on validator list btw 👍🏻
reward

cdc-Hitesh:
Awesome work.
Got some feedback.

- [ ]  Had to open withdraw all rewards two times to see the list
- [ ] The index count with Validators list can be removed to avoid confusion
- [ ] `Rewards` can be replaced with `Total rewards`

crypto-matto:
## Issue
- [ ] Hide Withdraw All Rewards button when reward list is empty
hide
hide2

src/pages/staking/staking.tsx
```diff
@@ -1788,6 +1925,31 @@ const StakingPage = () => {
                 : ''}
             </div>
           </div>
+
+          <ModalPopup
```
crypto-matto, line 1929: I'm seeing 2 `unbonding-modal`. Is this one duplicated and added without reason?

cdc-Hitesh (approved):

PR #1089, opened 2022-05-17
feat: revoke token permission

XinyuCRO:
Support revoke token permission on both Cronos main net and test net.

## Notice
- There is a cache issue on Cronos node RPCs, so for testing, when the confirm drawer shows, have to wait about half a minute to click the confirm button to prevent the `Wrong response id` issue

# Screenshot

image

image

image


crypto-matto:
Really admire the grouping~ 👍🏻 Yet it seems not working for me on my side? 🙈 
## Issue
- [ ] Grouping on permission under the same token is not working. 
asdfasdf

## Suggestion
- Isn't it better to put this `Revoke Permission` tab inside a new `Config` tab of `DApp Browser` instead? Like what you have implemented on Chain switch support. 
https://github.com/crypto-com/chain-desktop-wallet/pull/1032

XinyuCRO:
@crypto-matto 

> Grouping on permission under the same token is not working.

You can retry with a `yarn install`, it's a new feature in Ant Design

For the suggestion I will leave it as it is, will do refinments in later version after we can switch to different EVM chains

crypto-matto:
This is also thoughtful 👍🏻 
error

src/pages/settings/tabs/RevokePermission/RevokePermission.tsx
```diff
@@ -0,0 +1,137 @@
...
+            flexDirection: 'column',
+            alignItems: 'center',
+            justifyContent: 'center',
+          }}
+        >
+          <ErrorXmark />
+          <div>{error.toString()}</div>
+          <div>
+            <Button type="primary" onClick={() => fetch()}>
+              Retry
```
crypto-matto, line 93: Import `translation` at the top and change this 
```suggestion
              {t('general.retry')}
```

src/pages/settings/tabs/RevokePermission/CRC20TokenList.tsx
```diff
@@ -0,0 +1,399 @@
...
+  if (flatternedApprovalledData.length === 0) {
+    return (
+      <div
+        style={{
+          display: 'flex',
+          justifyContent: 'center',
+          alignItems: 'center',
+        }}
+      >
+        No Approvalled Data
```
crypto-matto, line 201: Add translation for `No Approval Data` / `No Approved Data`. May be use a light grey color & add more paddings to it. 
aas

src/pages/settings/tabs/RevokePermission/RevokePermission.tsx
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  if (
+    !cronosAsset ||
+    !cronosAsset.config?.nodeUrl ||
+    !cronosAsset.config.indexingUrl ||
+    !transferEvents ||
+    !approvalEvents
+  ) {
+    return <React.Fragment />;
```
crypto-matto, line 108: I think this one shall return something similar to `No Approval Data` / `No Approved Data`

crypto-matto:
I think once after the `Confirm` button is clicked, you could handle this modal by either: 
cancel
- Leaving the modal open, but disable both cancel & close button
- Close this modal directly

crypto-matto (dismissed):

crypto-matto (approved):

PR #1106, opened 2022-05-27
fix: gas form validation

XinyuCRO:
show a validation error if the gas related input numbers are negative

image


crypto-matto:
## Issue 
- [x] App crashes when for empty input

crypto-matto (approved):

PR #1114, opened 2022-05-31
Feature/#1093/backend

yi-cro:
1. **(Current PR)**
- [x] `Restake rewards from a SINGLE validator in a single tx`

--

2. **(To be done in the next PR)**
- [ ] `Restake rewards from MULTIPLE validators in a single tx`


src/service/TransactionSenderService.ts
```diff
@@ -375,6 +379,121 @@ export class TransactionSenderService {
...
+    const broadCastResult = await nodeRpc.broadcastTransaction(signedTxHex);
+    await Promise.all([
+      await this.txHistoryManager.fetchAndUpdateBalances(currentSession),
+      await this.txHistoryManager.fetchAndSaveDelegations(nodeRpc, currentSession),
+    ]);
+
+    return broadCastResult;
+  }
+
+  public async sendRestakeRewardsTransaction(
```
cdc-Hitesh, line 438: We can pull off these changes.

PR #1141, opened 2022-06-23
Issue/#949

yi-cro:
Problem: Show user voted proposals

- Voting History Page


crypto-matto:
## Issues
- [ ] The Voting History table is missing margin / padding
- [ ] Clicking the proposal item in voting history will not go into proposal detail immediately. I need to click `Back` in order to.

crypto-matto (approved):

PR #1164, opened 2022-06-30
chore: update cronos dapps

XinyuCRO:
TIP: we can use https://bulkresizephotos.com/zh-tw to compress images, it's very convenient, 10kb for each logo is good enough


crypto-matto:
I used tinyPNG before https://tinypng.com/

crypto-matto (approved):

PR #1167, opened 2022-07-04
Cosmos Hub Integration

crypto-matto:
## Background
We're looking into bringing Cosmos Hub Chain (ATOM) support on Asset Transfer with Ledger Support. The existing `chain-jslib` is not supporting other Cosmos chains natively. We're looking into integrate `cosmos/cosmjs`, which is the generic JS based library maintained by Cosmos. 

## Study
- Study the effort in integrating `cosmjs`
- Study the effort & possibility to support other Cosmos chains in `chain-jslib`

### Generic ATOM Support
- [x] Integrate `cosmjs` usage
- [x] ATOM asset transfer transaction
- [x] Generic Address Validator with different tendermint assets
- [x] Apply network configs by Tendermint assets instead of `currentSession`
- [x] Transaction History Indexing service migration
### Ledger Support
- [x] ATOM asset creation
- [x] Derivation path input
- [x] ATOM asset migration
- [x] ATOM asset Transfer transaction


XinyuCRO:
after an asset migration, there are some duplications of old assets
image

crypto-matto:
> after an asset migration, there are some duplications of old assets image

Thanks for the finding. It took me so much time to dig out what happened. But it's fixed

XinyuCRO (approved):

PR #1169, opened 2022-07-05
Issue/#1154/governance proposal frontend

yi-cro:
**Background**
We're looking into support Text Proposal submission on Desktop Wallet in Phase 1. In order to prevent spam proposal submissions, we enforce users to have an initial deposit of 1000 CRO from our client side.

**Tasks**
- [x]  Implement Governance Proposal Creation modal
- [x]  Form Validation on minimum 1000CRO initial deposit
- [x]  Integrate API signSubmitTextProposalTransaction() with frontend modal
- [x]  Implement Deposit Proposal workflow


crypto-matto:
## Issues
- [ ] Password Modal is not closed after a successful password entry
- [ ] The CRO amount input field will cause an app crash when the field is empty during number correction
- [ ] After the proposal submit, no tx is broadcasted
- [ ] A success / error modal is missing. People don't know if the broadcast is successful or not. 
- [ ] Leave more margin-right to the `Create Proposal` button
- [ ] The `Create Proposal` title is having a grey background, but a grey background should be at the description `It is recommended that you do the following before submitting a proposal:...` instead.  
## Suggestions
- Description may use a TextArea input instead

crypto-matto:
## Issue
- [ ] Typo `governance.modal2.form.proposalDeposit.wanring`. Also be aware that I see you may have deleted some items in the language spreadsheet, which leads to some translations are missing, e.g. `general.confirm`, `navbar.wallet.list`.
https://docs.google.com/spreadsheets/d/1tzmTGMyaex1JUv0HV1HQtisIOPOl3SpCoXIhXDlO_-U/edit#gid=0
Please double check if there're any other unintended deletion. 
- [ ] Even the validation doesn't pass, users are able to submit proposal, e.g. initial CRO with lower than minimum 1000 CRO requirement
- [ ] After the `Submit` button is clicked & loading, we shall disable the `Cancel` & `close` button. Pass `closable` props to `<ModalPopup>` and make good use of `confirmLoading` to achieve so. You may reference: `home.tsx` => `<ModalPopup>`
- [ ] Missing a grey background under `Create Proposal`
backg
bg

- [ ] See if you could align this margin-right
mar
marginr

crypto-matto:
## Issues
- [ ] There are still missing translations in places like placeholder. You may consider to remove placeholders, or put translation variables into it. 
err23

- [ ] This value is still typo `governance.modal2.form.proposalDeposit.wanring`. Please change to `governance.modal2.form.proposalDeposit.warning`

- [ ] Voting modal & Proposal modal got messed up. 
  - When I go to a voting stage proposal and click `Send Vote`, the `Create Proposal` modal is opened up instead. 
  - When I'm using a Ledger wallet and clicking `Create Proposal`, the `Send Vote` modal is opened up instead.

crypto-matto:
## Issues
### Deposit Modal
erc
May be it would be easier to implement another modal component, instead of using the same one. 

- [ ] Incorrect Title. Should be `Deposit`
- [ ] Field should be `Amount` instead of `Initial Deposit`
- [ ] The validation is able to bypass & goes right into the next step, even for scenarios like: insufficient balance, incorrect amount input, etc. 

### Password Modal
err2

- [ ] It shouldn't have 2 overlays appears at the same time. Should be only showing Password Modal & other modals shall not be visible. 
- [ ] After closing the Password Modal, it shall open the previous Deposit Modal. 

### Confirmation Modal
errr2
- [ ] This modal shall show up only after password validation succeeded. 
- [ ] When the deposit transaction is successful, a broadcast success modal `<SuccessModalPopup>` shall show up. The same goes for broadcast error modal `<ErrorModalPopup>`.

crypto-matto:
Admire your design on Proposal Type & the status of deposit visual 👍🏻 
des

crypto-matto:
By the way I just found out the weirdness on why your form validations are always bypassed. 

### Things to notice
- You're missing out passing an `onSubmit` props to `<Form>`
```
onFinish={() => {
                  allProps.setModalType('deposit');
                  setDepositModalVisible(false);
                  setConfirmDepositModalVisible(true);
                  setTimeout(() => {
                    showPasswordInput();
                  }, 200);
                }}
```
- Inside the submission button, for `onClick()` you should call `form.submit()`

Then the validation would work properly

### Reference
https://ant.design/components/form/

crypto-matto:
## Issue
- [ ] In `Create Proposal` Modal, the `Initial Deposit` field is not editable

crypto-matto:
I think this PR is finally ready 🥳 Will merge it once we roll out our next public release.

crypto-matto (approved):

LDbigG (approved):

LDbigG:
issue/#1154/governance-proposal-frontend

LDbigG (approved):

crypto-matto (approved):

Issue #1178, opened 2022-07-11
Merge ethereum-integration & its based branches into dev


crypto-matto:
## Branches
- cosmos-integration
- feat/wallet-connect

## Task
- [x] https://github.com/crypto-com/chain-desktop-wallet/pull/1163
- [x] https://github.com/crypto-com/chain-desktop-wallet/pull/1167
- [x] https://github.com/crypto-com/chain-desktop-wallet/pull/1188
- [x] https://github.com/crypto-com/chain-desktop-wallet/pull/912

PR #1188, opened 2022-07-13
feat: redirect users to etherscan to check eth related TXs

XinyuCRO:
[#1178](https://github.com/crypto-com/chain-desktop-wallet/issues/1178)

image


crypto-matto:
## New things added
- Make use of `renderExplorerUrl()`, switching explorer accordingly
- Apply the same measure on ATOM Testnet
- Fixed ATOM explorer configs

PR #1207, opened 2022-07-27
Fix: QA reported issues

crypto-matto:
## Issues
- [x] App crash on some address's voting history 
- [x] No back button for proposals entered from voting history
- [x] Sorting not working for voting history
- [x] Restake All rewards failure when > 10


Karam19925 (commented):
>>

Issue #1145, opened 2022-06-24
Add Notifications section


JJWW34:
## Study
- The behavior of Electron notification (https://www.electronjs.org/docs/latest/tutorial/notifications)
- Anything need to handle for silent mode in different OS (https://github.com/felixrieseberg/electron-notification-state)
- Whether we need a mailbox for users to revisit the notifications / messages or not
- When & where shall we push notifications to users
  - After any successful transactions
  - When token balances updated

## Task
- [x] https://github.com/crypto-com/chain-desktop-wallet/issues/1196
- [ ] Implement notifications in various necessary places
- [ ] Decide if we need an external endpoint to push notifications to users

calvinaco:
Mockup ETA: 19 July

Issue #1196, opened 2022-07-19
Notification section research & design


XinyuCRO:
## Design purpose

- Serve public notification as a static file to reduce the maintenance costs
- Design an in-app notification center that,
  - every module can donate notifications to the center
  - the center can manage remote public notifications and in-app notifications, and mix them properly 

## Detail

Separate notifications concept to local and remote

<img width="526" alt="image" src="https://user-images.githubusercontent.com/91446598/179882886-b3c8fcbb-abd6-472b-b069-a685094d03cc.png">

### Local Notification

Notifications based on the behavior of the current wallet users eg, transaction complete, response from customer service, they are all being delivered within the client

```ts
interface LocalNotification {
  content: string
  hasRead: bool
  icon: string // https://ant.design/components/icon/
  createdAt: Date
}
```

### Remote notifications

Contains the info we want to sync with all the Desktop Wallet users.

All remote notifications will be stored at GitHub as a static json file. Gist is ok, easy to update, no branch concepts, single source of truth, an example of the notification file: https://gist.github.com/XinyuCRO/8bb2405059681fdd6e2e2812a2c5aed6

```json
[
  {
    "id": 0,
    "content": "Scheduled upgrade ...",
    "created_at": 1658220227,
    "expire_at": 1660869827
  }
]
```

#### remote notification fetching & parsing
```mermaid
flowchart TD
    A[Fetch remote notifications] -- every 5 minutes --> A  -- for each notification item
    --> C{expired?}
    C -- No --> D{read?}
    D -- No --> E[push to notification center]
```

PR #1208, opened 2022-07-28
feat: Notification Center

XinyuCRO:
Closes https://github.com/crypto-com/chain-desktop-wallet/issues/1145

## Screenshot

image

image

## Notification Management

We use GitHub Repo to manage public notification. `config/notifications/dev.json` for testing and `config/notifications/release.json` for releasing notifications.

- dev endpoint: https://raw.githubusercontent.com/crypto-com/chain-desktop-wallet/dev/config/notifications/dev.json
- release endpoint: https://raw.githubusercontent.com/crypto-com/chain-desktop-wallet/dev/config/notifications/release.json

These two links point to the latest version of the notifications on the `dev` branch, which mean if we make changes to `config/notifications/release.json` and push it to the `dev` branch, we are pushing pubic notifications to users

Every time the client start and every 5 minutes, the client will fetch the endpoint, and determine whether there are new notifications, there is some cache-control on the endpoints, so there might be some slight delay

The detailed design can be seen here: https://github.com/crypto-com/chain-desktop-wallet/issues/1196


src/pages/notification/index.tsx
```diff
@@ -0,0 +1,161 @@
...
+                      size={48}
+                      icon={IconFromNotification(item)}
+                      style={{ backgroundColor: '#7B849B' }}
+                    />
+                  </div>
+                </div>
+                <div style={{ marginLeft: '10px', overflow: 'hidden' }}>
+                  <Typography.Paragraph
+                    style={{ fontSize: '14px' }}
+                    ellipsis={{ rows: 1, expandable: true }}
```
crypto-matto, line 103: May use `symbol` to input expand translation
```suggestion
                    ellipsis={{ rows: 1, expandable: true, symbol: t('general.notification.expand') }}
```

crypto-matto:
Is there any expire mechanism for out-dated CS notifications? 
螢幕截圖 2022-08-02 下午2 19 17
Also I noticed that the imported notification is based on the language I selected at the time. 
I'm just thinking if we should instead store the index of translation here, for standardized notification
```
    postLocalNotification({
      content: 'general.customerService.notification.body',
      type: 'customerService'
    })
```
Then decode here.
``` 
<Typography.Paragraph
      style={{ fontSize: '14px' }}
      ellipsis={{ rows: 1, expandable: true }}
>
    {t(item.content)}
</Typography.Paragraph>
```
But I also realize that this use case is not common, as not much people will switch languages in between. 
You may decide whether to make the adjustment or not.

XinyuCRO:
@crypto-matto 

>Is there any expire mechanism for out-dated CS notifications?

No

>Also I noticed that the imported notification is based on the language I selected at the time.

Yeah, not many cases, also users will get these notifications when they are using it, which means they can definitely read these different languages

crypto-matto:
> @crypto-matto
> 
> > Is there any expire mechanism for out-dated CS notifications?
> 
> No

May be in a long run, we may apply a removal of notifications, during the notification list fetching. When we got `isRead`: true & `createdAt` was like 7 / 14 or 30 days ago, then we remove the notification from the list. But guess it's too early for this & the list won't be too long for most people, as not many would use Live Chat too often. 

Let's merge this then :+1:

crypto-matto (dismissed):

Issue #1159, opened 2022-06-27
Feature: Integrate a Token Price chart in Token Detail Page


crypto-matto:
## Main Task
- #1157 
## Tasks
- [x] Locate and integrate a chart plotting library in visualizing the token price view
- [x] The chart shall support displaying token price in different selected time period
- [x] Use the default line chart layout provided by the library, with only adjusting the color into a bluish theme fitting our DeFi Desktop Wallet
- [x] A selection of time period button on the top-right corner above a chart: e.g. 8H,1D,1W,1M,3M,6M.
- [x] The chart will change accordingly when any time period button is clicked
- [x] Hovering any point along the line in the chart will display the price at that point
- [x] The price chart shall put inside under the Token Detail section

Optional: Display 24h price change in %

Issue #1158, opened 2022-06-27
Feature: Integrate with our Crypto.com Price API to fetch token prices over time


crypto-matto:
## Main Task
- #1157 

## Task
- Integrate with our Crypto.com Price API to fetch token prices market data over time: 8H,1D,1W,1M,3M,6M
- Implement a backend function in `MarketApi.ts`, supporting different assets & time period as parameters. 

### Price API
API endpoint to fetch token market prices over timestamp: 
```
https://price-api.crypto.com/price/v2/all/{slug}?convert={currency}
```
e.g. https://price-api.crypto.com/price/v2/all/ethereum?convert=USD

Issue #967, opened 2022-02-04
Bug: Retrieval of Ethereum Chain's token market prices on Cronos Tokens


mattopolitan:
## Issue
<img width="1175" alt="螢幕截圖 2022-02-04 下午9 57 34" src="https://user-images.githubusercontent.com/12154131/152541133-df9ff30a-3f74-4396-a4f6-1384ed209d38.png">
The existing token price retrieval mechanism for Cronos asset is mistakenly aligning token prices from Ethereum Chain which has the same token denom. 

## Solution
Crypto.com Price API team has added `cronos-ecosystem` into their tag return on https://price-api.crypto.com/price/v1/tokens/{slug}
Reference: https://price-api.crypto.com/price/v1/tokens/solana
```
"tags":["solana-ecosystem","alameda-research-ventures","andreessen-horowitz"]
```

For tokens on Cronos, it will be: 
```
"tags":["cronos-ecosystem"]
```
We'll be using this as an identifier. CRC20 tokens will only display token prices if `cronos-ecosystem` exists in the tag

Our approach:
1. `https://price-api.crypto.com/price/v2/all-tokens` to get all slug
  - Unfortunately `https://price-api.crypto.com/price/v2/all-tokens` weren't allow any filtering parameters apply. Which means all slugs will be returned.
2. `https://price-api.crypto.com/price/v1/tokens/{slug}` get a token market price data with the slug

PR #1212, opened 2022-08-04
feat: Price charts

### LINKED ISSUE #1157: Feature: Token Price chart
## Background
Currently we only reflect real time token prices. We're missing backend function to fetch token price changes over time & a visual chart tool to display it.

## Task
- [x] #1158 
- [x] #1159 

### Price API
API endpoint to fetch token market prices over timestamp: 
```
https://price-api.crypto.com/price/v2/all/{slug}?convert={currency}
```
e.g. https://price-api.crypto.com/price/v2/all/ethereum?convert=USD

### Visual Chart plotting tool
- Determine which visual tool library can be best fitted

XinyuCRO:
Closes #1157 #1158 #1159

image


src/pages/assets/components/ChartArea.tsx
```diff
@@ -0,0 +1,103 @@
...
+    getContainerSize();
+    window.addEventListener('resize', getContainerSize);
+    return () => window.removeEventListener('resize', getContainerSize);
+  }, []);
+
+  const fetchTokenPrices = useCallback(async () => {
+    setLoadingData(true);
+    try {
+      const data = await croMarketPriceApi.getTokenPrices(
+        `${asset.symbol}`,
```
crypto-matto, line 58:
```suggestion
        `${asset.mainnetSymbol}`,
```
Testnet tokens like TCRO is not fetching any data. Use `asset.mainnetSymbol` instead

crypto-matto:
### Issue
- #967 
```
getTokenPrices(cryptoSymbol: string, fiatCurrency: string, interval: string) {
... 
    const tokenSlugInfo = allTokensSlugMap?.find(tokenSlug => tokenSlug.symbol === cryptoSymbol);
    const slug = tokenSlugInfo?.slug;
... 
}
```
The problem with using barely `cryptoSymbol` is that it may fetch token prices of identical symbol tokens from other chains, resulting in an incorrect price chart. Therefore we have to identify the corresponding slug based on more criteria: 
1. For CRC20 tokens, check whether the API returned data `tags` includes `cronos-ecosystem`
2. For ERC20 tokens, sadly the above method doesn't work as many of them includes `ethereum-ecosystem`. I could only apply our whitelistedERC20Tokens whitelist table for filtering. 

You may reference on `getTokenPriceFromCryptoCom()`, see how I handled it. You may revise it and put the redundant logic into another function.

XinyuCRO:
if only they respond token list with contract addresses, the token slugs may be the same under the same chain

crypto-matto:
> if only they respond token list with contract addresses, the token slugs may be the same under the same chain

Ya sadly they don't. They don't provide the token contract address from any of their returned data neither. `Slug` is the only thing we got to locate the corresponding token.

crypto-matto:
## Updated
- Added Dynamic price figure
螢幕截圖 2022-08-04 下午5 35 49


- Default Round price figure to 2 decimals

crypto-matto (dismissed):

crypto-matto (approved):

PR #1223, opened 2022-08-16
Problem: Single bad actor operating multiple nodes

Terra854:
The nodes in the PR are all confirmed to be operated by a single entity and has used them to purposefully decommission and then immediately recommissioning the nodes multiple times, causing whoever is at the lowest in the validator set to have a significant hit in uptime whenever the node joins or leaves the active set.

Most of the active validators are in agreement that this should not continue and therefore, request that these validator address be blacklisted

### DIFF
--- a/config/app.moderation.json
+++ b/config/app.moderation.json
@@ -10,7 +10,12 @@
             "suspicious": [
                 "crocncl1xxuf6tsfsqm2zd2ype656dyr73x4fsjgd6m9en",
                 "crocncl1av95h6se2ptx0lsqkhwdy7u6eac2xusxrldl9d",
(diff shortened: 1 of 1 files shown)


Perseverance42:
Have to agree, especially the nodes which did unbind self-stake to get the next node into the active validator set, just to hopefully gain enough delegations on the new node to kick the 100th validator of the list by replenishing the self-stake on the original node. 
This behavior signals that they do not intend to operate seriously on the chain and showed malicious intent to other validators and especially to their delegators by purposefully putting delegators into jail for personal gain.

crypto-matto:
Hi @Terra854 @Perseverance42 , thanks for the reported list of addresses. 

We’ve added them into our suspicious list. Further investigation has to be done. Once they're determined to be blacklisted, we'll proceed with the PR.

PR #1235, opened 2022-09-05
Fix: Missing addChain support inside most of the DApps

crypto-matto:
## Issues
- [x] The add chain functionality is breaking in almost all the dApps. Normally, DApps will not try to call `wallet_addEthereumChain` directly. Instead, DApps will call `wallet_addEthereumChain` when it failed to switch to that chain. 
  - When switch chain error happens, throw an error with error.code set to `4902`
  - Reference: https://docs.metamask.io/guide/rpc-api.html#usage-with-wallet-switchethereumchain
- [x] Some DApps doesn't include Explorer URL in adding chain config, which results our app crashes
- [x] Incorrect Network ID format from `preload.js`, causing some DApp crashes, e.g. `https://app.ebisusbay.com/`


src/pages/dapp/browser/preload.js
```diff
@@ -488,6 +488,16 @@ class Web3Provider extends EventEmitter {
    */
   postMessage(handler, id, data) {
     if (this.ready || handler === 'requestAccounts') {
+      if (handler === 'switchEthereumChain') {
```
XinyuCRO, line 491: postMessage is not a good place to handle business logic, and the message is not always failed on this handler

can send the error code back in this line: https://github.com/crypto-com/chain-desktop-wallet/blob/dev/src/pages/dapp/browser/useIPCProvider.ts#L532

XinyuCRO (changes requested):
Haven't encountered many problems like this, I remembered many dapps I tested before won't check the error code, but it's a good practice to follow the MetaMask

PR #1242, opened 2022-09-13
Revert "Update Package v1.3.1"

LDbigG:
Reverts crypto-com/chain-desktop-wallet#1241


LDbigG (commented):
LDbigG:revert-1241-dev

PR #1252, opened 2022-09-27
Update Package v1.3.2

crypto-matto:
### Additions
- Optionalize import of Cosmos Hub ATOM in Ledger wallet
- Revise migration flow for new assets
- Rebrand Cronos Chain
### Bug Fixes
- App crash when Cosmos Hub ATOM not migrated in Bridge Transfer
- Invalid Gas Token on various Tendermint related operations


config/notifications/release.json
```diff
@@ -1 +1,6 @@
-[]
+[
+  {
+    "id": 1,
+    "content": "Crypto.com DeFi Desktop Wallet v1.3.2 is released!"
```
XinyuCRO, line 4: add `created_at` and `expire_at`, can refer to dev.json

XinyuCRO (commented):
Will use notification to notify update conflict with the electron native update? might be too many notifications for users

config/notifications/release.json
crypto-matto: > Will use notification to notify update conflict with the electron native update? might be too many notifications for users

True 🤔 Let me remove it & I'll keep the `master` branch `release.json` instead.

PR #1273, opened 2022-11-03
Problem: Fix non-hex message signing through walletConnect

crypto-matto:
## Issue
Message signing should often comes in Hex
螢幕截圖 2022-11-03 下午5 43 01
But for some scenario, when a non-Hex string pass to WalletConnect, spaces got trimmed: 
image (2)
image

Tracing back to the source, it's the `WalletConnect` Class from lib `@walletconnect` causing the issue.

### DIFF
--- a/src/service/walletconnect/useWalletConnect.ts
+++ b/src/service/walletconnect/useWalletConnect.ts
@@ -2,7 +2,7 @@ import WalletConnect from '@walletconnect/client';
 import { IJsonRpcRequest } from '@walletconnect/types';
 import { useCallback, useState } from 'react';
 import { useRecoilState } from 'recoil';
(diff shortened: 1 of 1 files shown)


XinyuCRO:
I think it's not a issue, DApp should follow the spec and pass the hexed string instead

XinyuCRO (approved):

crypto-matto:
> I think it's not a issue, DApp should follow the spec and pass the hexed string instead

Agree. So far I also see DApps are usually passing hex string. I'll update this as a safety measure though, as we see DeFi mobile app also got handled this. I've reflected this to @damoncro to consider hexing the string before passing the message for his service, as not sure if some other wallets will miss the handling of this as well~

Issue #1274, opened 2022-11-04
Problem: WalletConnect not working on Windows


crypto-matto:
## Issue
When clicking the WalletConnect from the DApp, it always opens a new Desktop Wallet app (then closes quickly) instead of connecting to the original opened one. This issue only appears on Windows. 

The WalletConnect path is universal on all OS: 
```
cryptowallet://wc?uri=wc%3Aeaa781bf-41ee-4de8-b384-6be23c7a7c62%401%3Fbridge%3Dhttps%253A%252F%252Fw.bridge.walletconnect.org%26key%3Dbac7e125a9ab1293fed226bc79d53e1c98ffa2f4b7eaef1c820e150f99883969
```

PR #1276, opened 2022-11-07
fix: WalletConnect not working on Windows #1274

XinyuCRO:


crypto-matto:
When choosing WalletConnect on Windows, there's an error when I'm trying it with ` yarn run electron:windev`. Build binary is also experiencing the same issue.
螢幕截圖 2022-11-07 下午3 16 18

PR #1379, opened 2023-09-19
fix: only support http/https protocol in the dapp browser

XinyuCRO:
Background: https://www.cybersecurity-help.cz/vdb/SB2023091322

Solutions from: https://github.com/electron/electron/issues/1378#issuecomment-265207386

- Only allow HTTP and HTTPS protocol, and block others
- `isValidURL` is a method copied from `src` folder, seems no way to share code between main and renderer process, so just make a simple duplication here


crypto-matto (approved):
LGTM. Able to block non-http protocols. While haven't received Security Team's reply though, let's prepare a new internal release first.

PR #1401, opened 2024-01-15
Fix: New Window Handling

crypto-matto:
## Background
Test hosting a script as below: (opening the calculator app on Windows)
```
<script>
   window.open("calculator:")
</script>
```

The app will pops up with the Calculator app. 

### Attempts to fix
The above script will trigger: 
- `webview` => `new-window`
- `window` => `will-navigate`

Both navigating `url` to `calculator:`

Therefore, tried to perform the below measures:
- Blocks any new windows from being opened, with event `new-window` 
- Capture `window` => `will-navigate` event

While no more windows will open, yet it is not able to disable the execution on `window.open` by any means. 

Setting `javascript` to `false` is also not able to stop it from executing
```
if (contents.getType() == 'webview') {
    // blocks any new windows from being opened
    contents.on('new-window', (event, url, frameName, disposition, options) => {
      console.log('new-window', url, frameName, disposition, options)
      options.webPreferences = {
        ...options.webPreferences,
        javascript: false,
      };
      event.preventDefault();
      require('electron').shell.openExternal(url);
      return;
    })
```

### Reference 
- https://shabarkin.medium.com/1-click-rce-in-electron-applications-79b52e1fe8b8
- https://benjamin-altpeter.de/shell-openexternal-dangers/


XinyuCRO:
Fixed, can use https://cly5ry-1234.csb.app/ for testing, it will open smb:// on load, but now DApp browser will prevent opening it.

PR #1418, opened 2024-03-05
Fix: Update windows cert

crypto-matto:
## Background
A Windows Cert signing release has been suspended for some times due to the expired cert. Now that the cert is renewed. We're able to deliver signed Windows release again.


crypto-matto:
Fixed in #1455

